PR #16, opened 2026-07-12
feat: startup validation fails on declared-but-unimplemented plugin stubs

guyash1:
## Summary
Per the suggestion from @asaf — validate at startup that every plugin function the YAML declares was actually implemented, instead of letting forgotten `agentctl generate` stubs explode one-by-one mid-request against real users.

```
✗ [tools.get_order_status] Tool 'get_order_status' is declared but not implemented (generated stub): plugins/tools/get_order_status.py
✗ [planner.resolvers.user_name] Resolver 'user_name' is declared but not implemented (generated stub)
✗ [plugins.access] AccessResolver.can_access is not implemented (generated stub) — every protected node will be hidden for every caller
```

The failure meets the developer at startup — never the user at runtime. (Complements #9's runtime fail-closed: loud at boot, quiet in production.)

## How it works
- **Static only:** plugin files are parsed with `ast`, never executed — zero side effects. A stub is a body that is exactly `raise NotImplementedError` (± docstring).
- **Conservative:** only positively identified stubs and missing files are reported. Dynamic definitions, methods inherited from bases other than the conventional `shared.py`, and unparseable files are skipped (runtime loaders diagnose those better).
- **One choke point:** wired into `SystemSpecValidator`, so `agentctl validate/run/chat` and both API servers all get it.
- **Coverage:** tools, per-agent + shared resolvers, `AccessResolver.can_access` (only when protected nodes exist), and `mcp_auth` `get_headers` (only when `auth: true`).

## Found real cases immediately
Both custom tools in the flagship example (`generate_decision_matrix`, `build_learning_plan`) were still generated stubs — the new scan caught them. Implemented both as small deterministic demo tools so the example validates and actually works end-to-end.

## Files changed
- `src/agent_engine/core/plugin_stubs.py` — new scan module
- `src/agent_engine/core/validator.py` — 2-line wiring
- `tests/core/test_plugin_stub_scan.py` — 15 tests (stub/implemented/missing per plugin kind, shared-base inheritance both ways, dedupe, no-false-positive on unknown bases, validator wiring)
- `examples/enterprise-knowledge-assistant/plugins/tools/*` — demo implementations

## Architecture rules respected
Validate-before-compile layering preserved; scan lives in `core`, no runtime changes; YAML stays declarative; nothing executed during validation.

## Commands run
`make check` — pass (ruff, mypy, **367 tests passed**, 15 new). Flagship example: `agentctl validate` ✓.

## Out of scope / follow-ups
- Hook plugin methods (need `plugins.toml` resolution) — happy to add as a follow-up.
- If you'd prefer stubs to be a *warning* rather than an error in some dev flows, the severity decision is one line — say the word.

### DIFF
--- a/examples/enterprise-knowledge-assistant/plugins/tools/build_learning_plan.py
+++ b/examples/enterprise-knowledge-assistant/plugins/tools/build_learning_plan.py
@@ -1,4 +1,26 @@
 def build_learning_plan(input: dict) -> str:
-    """Build a personalized learning roadmap from the collected research findings.
-"""
-    raise NotImplementedError
+    """Build a personalized learning roadmap from the collected research
+    findings.
+
+    Expects ``topics`` (ordered list of things to learn) and optionally
+    ``experience_level`` and ``weeks`` (total duration). Returns a phased
+    markdown roadmap the learning planner presents to the user.
+    """
+    topics = [str(t) for t in input.get("topics") or []]
+    if not topics:
+        return "No topics provided. Call with {'topics': [...], 'weeks': 4}."
+
+    level = str(input.get("experience_level") or "beginner")
+    weeks = max(int(input.get("weeks") or len(topics)), 1)
+    per_topic = max(weeks // len(topics), 1)
+
+    lines = [f"# Learning plan ({level}, ~{weeks} weeks)", ""]
+    week = 1
+    for topic in topics:
+        end = min(week + per_topic - 1, weeks)
+        span = f"Week {week}" if end == week else f"Weeks {week}-{end}"
+        lines.append(f"- **{span}:** {topic}")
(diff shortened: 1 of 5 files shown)


guyash1:
@Asaf-prog @AmitAvital1 review requested 🙏 (couldn't add you as formal reviewers from a fork — no triage permission.)

Quick review guide:
- Core logic: `src/agent_engine/core/plugin_stubs.py` (static AST scan, nothing executed)
- Wiring: 2 lines in `SystemSpecValidator`
- The scan caught two real stubs in the flagship example's tools — implemented as demo tools in this PR
- Two open decisions flagged in the description: hook-method scanning as follow-up, and error-vs-warning severity

PR #17, opened 2026-07-14
feat(models): openai-compatible vendor presets (zai, deepseek, moonshot, groq, xai, openrouter)

fjgbue:
The openai provider was hardcoded to api.openai.com. First pass just added base_url and api_key_env so provider: openai could point anywhere, but that means typing out the endpoint and key var by hand every time, even for vendors this table already knows about. This adds the actual integration on top of that: agent_engine/models/presets.py is a small registry mapping provider ids (zai, deepseek, moonshot, groq, xai, openrouter) to their base_url/api_key_env, resolved in build_chat_model before it dispatches to _build_openai_model. No new provider kind, no new dependency, it's ChatOpenAI either way.

model:
  provider: zai
  name: glm-5.2

That's the whole config now for a listed vendor. api_key_env still gets checked (ZAI_API_KEY in this case), it's just not typed into the YAML anymore. base_url/api_key_env still work directly, both as the escape hatch for anything not in the table (a different vendor, self-hosted Ollama/vLLM) and as an override on a listed preset, e.g. routing zai through an internal proxy without giving up the shorthand:

model:
  provider: zai
  name: glm-5.2
  base_url: https://llm-proxy.internal.example.com/zai
  api_key_env: INTERNAL_PROXY_KEY

Also fixed ARCHITECTURE.md and ROADMAP.md, which still only mentioned Anthropic and Bedrock even though Gemini and OpenAI have been in factory.py for a while, and updated both again now that there's an actual preset registry instead of just two generic fields.

Tests: 379 passing (up from 370), including a parametrized case that checks every preset resolves its own base_url/key correctly, one confirming an explicit base_url/api_key_env override wins over the preset, one for case-insensitive provider ids, and one confirming the unsupported-provider error message lists the preset ids alongside anthropic/bedrock/gemini/openai. ruff and mypy clean on every file touched.

### DIFF
--- a/src/agent_engine/models/factory.py
+++ b/src/agent_engine/models/factory.py
@@ -34,6 +34,8 @@ def build_chat_model(
     region: str | None = None,
     max_tokens: int | None = None,
     top_p: float | None = None,
+    base_url: str | None = None,
+    api_key_env: str | None = None,
 ) -> BaseChatModel:
     """Construct a chat model from flat config fields.
 
@@ -81,6 +83,8 @@ def build_chat_model(
             temperature=temperature,
             max_tokens=max_tokens,
             top_p=top_p,
+            base_url=base_url,
+            api_key_env=api_key_env,
         )
     supported = ", ".join(_SUPPORTED_PROVIDERS)
     raise ModelConfigurationError(
@@ -205,11 +209,19 @@ def _build_openai_model(
     temperature: float | None,
     max_tokens: int | None,
     top_p: float | None,
+    base_url: str | None = None,
+    api_key_env: str | None = None,
 ) -> BaseChatModel:
-    api_key = _resolve_openai_api_key()
+    # `api_key_env` lets this provider point at ANY OpenAI-compatible chat
+    # completions endpoint (Z.AI, DeepSeek, Moonshot, Groq, xAI, OpenRouter,
+    # a local Ollama/vLLM server, ...) by naming the env var that holds that
+    # vendor's key. Defaults to OPENAI_API_KEY to keep existing YAML working
+    # unchanged.
+    key_env_var = api_key_env.strip() if api_key_env and api_key_env.strip() else "OPENAI_API_KEY"
+    api_key = _resolve_openai_api_key(key_env_var)
     if not api_key:
         raise ModelConfigurationError(
-            "OpenAI provider requires OPENAI_API_KEY. "
+            f"OpenAI-compatible provider requires {key_env_var}. "
             "Set it in your environment before running agentctl."
         )
 
@@ -223,14 +235,17 @@ def _build_openai_model(
         ) from exc
 
     # OpenAI's chat model uses the same `max_tokens` name as extra's config, so
-    # no remapping is needed. The API key is passed explicitly so OPENAI_API_KEY
-    # is honored regardless of how the installed SDK resolves credentials.
+    # no remapping is needed. The API key is passed explicitly so the resolved
+    # env var is honored regardless of how the installed SDK resolves
+    # credentials. `base_url` repoints the client at any OpenAI-compatible
+    # endpoint; when unset, ChatOpenAI's own default (api.openai.com) applies.
     try:
         return ChatOpenAI(
             **_without_none(
                 {
                     "model": name,
                     "api_key": api_key,
+                    "base_url": base_url,
                     "temperature": temperature,
                     "max_tokens": max_tokens,
                     "top_p": top_p,
@@ -239,8 +254,8 @@ def _build_openai_model(
         )
     except Exception as exc:
         raise ModelConfigurationError(
-            "Could not initialize OpenAI chat model. Verify OPENAI_API_KEY is valid "
-            "and the model name is an OpenAI model your key can access."
+            f"Could not initialize OpenAI-compatible chat model. Verify {key_env_var} is "
+            "valid and the model name is one your key/endpoint can access."
         ) from exc
 
 
@@ -257,8 +272,8 @@ def _resolve_gemini_api_key() -> str | None:
     return key.strip() if key and key.strip() else None
 
 
-def _resolve_openai_api_key() -> str | None:
-    key = os.getenv("OPENAI_API_KEY")
+def _resolve_openai_api_key(env_var: str = "OPENAI_API_KEY") -> str | None:
+    key = os.getenv(env_var)
     return key.strip() if key and key.strip() else None
 
 
--- a/src/agent_engine/models/factory.py
+++ b/src/agent_engine/models/factory.py
@@ -16,6 +16,7 @@ from langchain.chat_models import init_chat_model
 from langchain_core.language_models import BaseChatModel
 
 from agent_engine.logging_config import log
+from agent_engine.models.presets import OPENAI_COMPAT_PRESETS
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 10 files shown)


src/agent_engine/models/factory.py
```diff
@@ -205,11 +224,19 @@ def _build_openai_model(
...
+    base_url: str | None = None,
+    api_key_env: str | None = None,
 ) -> BaseChatModel:
-    api_key = _resolve_openai_api_key()
+    # `api_key_env` lets this provider point at ANY OpenAI-compatible chat
+    # completions endpoint (Z.AI, DeepSeek, Moonshot, Groq, xAI, OpenRouter,
+    # a local Ollama/vLLM server, ...) by naming the env var that holds that
+    # vendor's key. Defaults to OPENAI_API_KEY to keep existing YAML working
+    # unchanged.
+    key_env_var = api_key_env.strip() if api_key_env and api_key_env.strip() else "OPENAI_API_KEY"
```
AmitAvital1, line 235: If api_key_env is "" we not should error so user meant to override and not use OPEN_API_KEY?

src/agent_engine/models/factory.py
```diff
@@ -16,6 +16,7 @@
 from langchain_core.language_models import BaseChatModel
```
AmitAvital1, line 16: I think you can remove comments, code are explained.
Comment should be only where the code its not explained or we are doing something need to explain

src/agent_engine/models/presets.py
```diff
@@ -0,0 +1,70 @@
+"""Known OpenAI-compatible vendor presets.
```
AmitAvital1, line 1: I'm not sure this file is needed. what you have built, its generic infra for custom models.
This led us to maintain code and each changes of the providers. let user declare api key name and urls and models, they write once time yaml and use model name each place

src/agent_engine/models/factory.py
```diff
@@ -44,11 +47,25 @@ def build_chat_model(
...
+    # filled in. YAML-supplied base_url/api_key_env still win if both are
+    # set explicitly alongside a preset id, so a preset can be overridden
+    # (a proxy in front of it, a different key variable name) without
+    # switching away from the shorthand.
+    preset_id = normalized_provider
+    preset = OPENAI_COMPAT_PRESETS.get(preset_id)
+    if preset is not None:
+        base_url = base_url or preset.base_url
+        api_key_env = api_key_env or preset.api_key_env
+        normalized_provider = "openai"
```
AmitAvital1, line 61: if we remove preset always provider will be openai

src/agent_engine/models/factory.py
```diff
@@ -44,11 +47,25 @@ def build_chat_model(
     if not model_name:
         raise ModelConfigurationError("Model name must not be empty.")
 
+    # A preset id (zai, deepseek, moonshot, groq, xai, openrouter, ...) is
```
AmitAvital1, line 50: what about checking if provider is gemini, but user declare base url or api key, or just one of them? we should error him on validate yaml that its not acceptable

AmitAvital1 (changes requested):

src/agent_engine/models/factory.py
fjgbue: empty string should error, yeah. 

the OPENAI_API_KEY default is meant for when the field is omitted, not when someone explicitly blanks it. An explicit "" means they were mid-override and got it wrong, and falling back silently just turns that into a confusing auth failure further down. I'll split the two: omitted stays defaulted, empty or whitespace raises. what u think?

src/agent_engine/models/presets.py
fjgbue: the maintenance worry is fair but smaller than it looks I think

these base URLs don't move, and the table is six entries. The generic base_url/api_key_env path already covers everything the registry doesn't, including overriding a preset, so the registry adds no capability, it just saves every user re-typing the same two known values for a vendor the library could already name, or at least that was my initial tought. But if the team would rather ship zero vendor knowledge though, killing presets.py is one commit since the generic path stands on its own, lmk

src/agent_engine/models/factory.py
fjgbue: yes, right, drop the registry and this branch goes with it, proider is just openai plus whatever base_url/key the user passed. they live or die together

fjgbue (commented):
Awaiting some further comments in the answers!

src/agent_engine/models/presets.py
AmitAvital1: @Asaf-prog wdyt? we need more opinion

src/agent_engine/parsers/yaml/parser.py
```diff
@@ -349,6 +350,17 @@ def _validate_model(path: str, raw: Any, errors: list[ValidationError]) -> None:
...
         )
 
+    if provider in _FIXED_ENDPOINT_PROVIDERS:
+        for field_name in ("base_url", "api_key_env"):
+            if raw.get(field_name) is not None:
+                errors.append(
+                    ValidationError(
+                        f"{path}.{field_name}",
+                        f"Not supported for provider '{provider}'; only OpenAI-compatible "
+                        "providers accept a custom endpoint or key variable.",
```
AmitAvital1, line 360: on error message: base url or api key naming variable?

src/agent_engine/parsers/yaml/parser.py
```diff
@@ -349,6 +350,17 @@ def _validate_model(path: str, raw: Any, errors: list[ValidationError]) -> None:
             )
         )
 
+    if provider in _FIXED_ENDPOINT_PROVIDERS:
```
AmitAvital1, line 353: instead of nested ifs lets try make it better like (pseudo) :
for field_name in ("base_url", "api_key_env"):
 if raw.get(field_name) is not None && provider in _FIXED_ENDPOINT_PROVIDERS
 return error

src/agent_engine/models/factory.py
AmitAvital1: I still wondering if user will declare on yaml api_key_env = "   " 
we will fall back to open api key default. but, he tried to override and maybe typo, so user experience in my view should error him.
Now we fallback

src/agent_engine/models/presets.py
AmitAvital1: Secondly i feel good with that. But we have to doc it also on official docs mdx files. focused and clear

AmitAvital1:
https://docs.extra-ai.co/docs/yaml-spec#supported-providers

We should documented this new feature.
Now we are doing supported providers on yaml refernce. i think we should add new section under features calling providers. detailed the pre defined providers, and supported open ai api providers and custom providers and how defining them

and here just keep example but add reference to this new section https://docs.extra-ai.co/docs/yaml-spec#supported-providers

AmitAvital1:
Hi, want share with you so i checked and our engine are supporting by that using langchain open api support by using this env var

like ollama 
OPENAI_API_KEY=ollama
OPENAI_BASE_URL=http://host.docker.internal:11434/v1

open router
OPENAI_API_KEY=sk-*****
OPENAI_BASE_URL=https://openrouter.ai/api/v1

we just put provider in 
defaults:
  model:
    provider: openai
    name: nvidia/llama-nemotron-rerank-vl-1b-v2:free
    temperature: 0.0
    
   
   so lets think how its affected since, we currently support it. but need to check how its works, and if we decide go with your fix make sure we have no collisions

PR #21, opened 2026-07-17
feat: scaffold missing prompt files during generate and validate them at startup (#18)

### LINKED ISSUE #18: generate doesn't scaffold missing prompt files declared in YAML
## Background

Prompt files are declared in `agents.yml` by relative path:

```yaml
agents:
  orders_agent:
    prompts:
      system: "prompts/orders/system.md"
```

`agentctl generate` (`src/agent_engine/generate/generator.py`) already scaffolds every other artifact a spec declares but doesn't yet have on disk — tool stubs, resolver stubs, `plugins/access.py`, MCP auth stubs, hook stubs — except prompt files.

## Current behavior

If `prompts.system: "prompts/orders/system.md"` doesn't exist yet, `generate` does nothing. The first sign of trouble is `SystemSpecValidator` (`src/agent_engine/core/validator.py:30-50`) failing `validate`/`serve` with:

```
Prompt file not found: prompts/orders/system.md
```

A hard stop, no scaffold offered — inconsistent with every other declared-but-missing artifact.

## Goal

1. `generate` creates any missing declared prompt file (`system` / `user` / `orchestrator`), matching the "create if missing, never overwrite" rule already used for every other stub type.
2. The generated file's body is a fixed sentinel line, e.g. `<!-- STUB: fill in this prompt -->` — the same role `raise NotImplementedError` plays for tool/resolver stubs.
3. Extend the existing stub scanner (`src/agent_engine/core/plugin_stubs.py`, already wired into `SystemSpecValidator`) to flag that sentinel the same way it flags unimplemented tool/resolver/access stubs today, so a forgotten prompt is caught at startup validation, not silently sent to the model.

rishu685:
Closes #18

### Background
Previously, prompt files declared in the system specification YAML were not scaffolded by `agentctl generate` when missing. This resulted in `SystemSpecValidator` failing with a hard "Prompt file not found" error during startup without offering a stub, unlike tools, resolvers, and hooks.

### Changes

1. Scaffold Missing Prompts (`src/agent_engine/generate/generator.py`):
   - Walks the spec graph and collects all declared prompt paths (`system`, `user`, and `orchestrator`).
   - Creates parent directories and writes missing prompt files using the sentinel content `<!-- STUB: fill in this prompt -->` under the "create if missing, never overwrite" rule.
   - Deduplicates prompt paths so shared templates are only written and reported once.

2. Validate Unimplemented Stubs (`src/agent_engine/core/plugin_stubs.py`):
   - Extended the startup scanner to inspect the contents of declared prompt files.
   - Flags prompt files containing the sentinel string as unimplemented, matching the style of other stub validation errors.
   - Safely ignores unreadable files (leaving them for the runtime loader to report) and correctly skips validation on files that don't exist yet to avoid double-reporting.

3. Tests (`tests/`):
   - Added generator tests verifying stub creation, idempotency (existing files are untouched), and deduplication.
   - Added scanner tests ensuring prompt stubs for all fields (`system`, `user`, `orchestrator`) are caught at startup.

### DIFF
--- a/src/agent_engine/core/plugin_stubs.py
+++ b/src/agent_engine/core/plugin_stubs.py
@@ -12,6 +12,8 @@ conservative — it reports only what it can positively identify:
 
 - a function/method whose body is just ``raise NotImplementedError``
   (optionally preceded by a docstring) is a stub;
+- a prompt file whose content is the sentinel ``<!-- STUB: fill in this prompt -->``
+  is a stub;
 - a missing plugin file for something the spec declares is an error;
 - anything else (dynamic definitions, methods inherited from bases other
   than the conventional ``shared.py``, unparseable files) is skipped and
@@ -51,6 +53,7 @@ class _Scanner:
         self._base_dir = base_dir
         self._seen_tools: set[str] = set()
         self._seen_mcp_auth: set[str] = set()
+        self._seen_prompts: set[str] = set()
         self.has_protected = False
         self.errors: list[ValidationError] = []
 
@@ -61,6 +64,7 @@ class _Scanner:
             self._check_tools(node.node)
             self._check_resolvers(node.node)
             self._check_mcp_auth(node.node)
+        self._check_prompts(node)
         for child in node.children:
             self.walk(child)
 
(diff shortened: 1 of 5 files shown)


src/agent_engine/generate/generator.py
```diff
@@ -89,6 +89,13 @@ def generate(self, spec: SystemSpec, base_dir: Path) -> GenerateResult:
                     result,
                 )
 
+        # Generate missing prompt stubs
```
AmitAvital1: not need this comment, code are explained

src/agent_engine/core/plugin_stubs.py
```diff
@@ -129,6 +133,33 @@ def _check_mcp_auth(self, agent: AgentSpec) -> None:
...
+        prompts = node.node.get_prompts()
+        for field_name in ("system", "user", "orchestrator"):
+            path_str = getattr(prompts, field_name, None)
+            if not path_str or path_str in self._seen_prompts:
+                continue
+            self._seen_prompts.add(path_str)
+
+            path = self._base_dir / path_str
+            if not path.is_file():
+                # Missing file is already reported by SystemSpecValidator._validate_node
```
AmitAvital1: I think we still need to return error here and not ignore this case

src/agent_engine/generate/generator.py
```diff
@@ -89,6 +89,13 @@ def generate(self, spec: SystemSpec, base_dir: Path) -> GenerateResult:
...
+        prompt_paths = _collect_prompts(spec.graph)
+        for path_str in prompt_paths:
+            path = base_dir / path_str
+            path.parent.mkdir(parents=True, exist_ok=True)
+            self._write(path, "<!-- STUB: fill in this prompt -->\n", result)
```
AmitAvital1: write are printing only file name, now we should maybe consider refactor write, maybe other commit so here we will print all path? because its not let us understand able which files really generates and location. WDYT?

AmitAvital1:
please add commit description as well :)

src/agent_engine/generate/generator.py
rishu685: hey,thanks for pointing out this! I completely agree!

I've updated the Generator to output relative paths from the base directory (e.g. create prompts/flights/system.md and create plugins/tools/lookup.py) instead of just the filenames. This makes it clear exactly where the files are being generated, and I've updated the test assertions to match.

AmitAvital1 (approved):

PR #30, opened 2026-07-18
feat: support fallback models and automatic retries on LLM failure (#25)

### LINKED ISSUE #25: Support a fallback model in YAML — retry on a backup model when the primary fails
## Background

A node's model is defined once in `agents.yml` — provider, name, and generation params. If the configured provider or model is unavailable (outage, rate limit, deprecated model), the request just fails today. There's no way to declare a backup.

## Purpose

Let a node declare a fallback model alongside its primary one. If the primary model fails, the request is retried against the fallback instead of failing outright.

## Proposed YAML

```yaml
model:
  provider: anthropic
  name: claude-sonnet-4-6
  fallback:
    provider: openai
    name: gpt-4o
```

`fallback` nests a model definition, same shape and same fields as `model` itself. It should be usable anywhere a model is defined today — per-agent, per-orchestrator, or in `defaults.model` — with no different behavior depending on where it's declared.

## Scope for v1

- Exactly one fallback model — no chains. A `fallback` block that itself declares a nested `fallback` is a validation error, not silently ignored.
- Fallback triggers on any failure from the primary, not narrowed to specific error types (rate limits, timeouts, 5xx) — simplest starting point, can be scoped down later if that proves too broad.

rishu685:
Closes #25

### Background
Currently, if a primary model fails due to provider outages, rate limiting, or connection drops, the agent/orchestrator execution fails immediately. To make the platform resilient, we need support for fallback model configurations.

### Changes
1. Spec & Parser Update:
   - Modified `ModelConfig` to include an optional nested `fallback: ModelConfig` field.
   - Updated `YAMLParser` to recursively parse `fallback` configurations.
   - Added validation in `_validate_model` to enforce exactly one level of nested fallback (strictly rejecting fallback chains/nested fallbacks under fallback models) and throwing a `ValidationError`.

2. Execution Engine Integration:
   - Updated `_build_model` to bind tools to the primary and fallback models individually before wrapping them using LangChain's `.with_fallbacks()` capability.
   - Triggers the fallback backup model on any model invocation or streaming exception.

3. Diagnostics & Inspect Output:
   - Updated the `agentctl inspect` command to display fallback model configurations for agents in offline inspection summaries.

4. Tests:
   - Added validation unit tests in `tests/cli/test_validate_command.py` checking validation checks on valid and invalid fallback configurations.
   - Added integration/E2E tests in `tests/engine/test_engine_flow.py` checking that both normal execution (`ainvoke`) and streaming (`astream`) gracefully recover when the primary model fails.
   - Added CLI inspect tests in `tests/cli/test_inspect_command.py` checking fallback output rendering.

### DIFF
--- a/src/agent_engine/core/spec.py
+++ b/src/agent_engine/core/spec.py
@@ -7,14 +7,18 @@ from agent_engine.core.execution import ExecutionPolicy
 
 
 @dataclass(frozen=True)
-class ModelConfig:
+class BaseModelConfig:
     provider: str
     name: str
     temperature: float | None = None
     region: str | None = None
     max_tokens: int | None = None
     top_p: float | None = None
-    fallback: ModelConfig | None = None
+
+
+@dataclass(frozen=True)
+class ModelConfig(BaseModelConfig):
+    fallback: BaseModelConfig | None = None
 
 
 @dataclass(frozen=True)
--- a/src/agent_engine/core/spec.py
+++ b/src/agent_engine/core/spec.py
@@ -14,6 +14,7 @@ class ModelConfig:
     region: str | None = None
     max_tokens: int | None = None
     top_p: float | None = None
+    fallback: ModelConfig | None = None
 
 
 @dataclass(frozen=True)
--- a/src/agent_engine/engine/langgraph/engine.py
+++ b/src/agent_engine/engine/langgraph/engine.py
@@ -31,8 +31,14 @@ from agent_engine.approvals.session_store import (
     SessionApprovalStore,
 )
 from agent_engine.core.execution import ExecutionPolicy
-from agent_engine.core.spec import AgentSpec, GraphNode, OrchestratorSpec, SystemSpec
-from agent_engine.core.spec import BaseModelConfig, ModelConfig as NodeModelConfig
+from agent_engine.core.spec import (
+    AgentSpec,
+    BaseModelConfig,
+    GraphNode,
+    OrchestratorSpec,
+    SystemSpec,
+)
+from agent_engine.core.spec import ModelConfig as NodeModelConfig
 from agent_engine.engine.engine import Engine
 from agent_engine.engine.langgraph.approval_provider import InterruptApprovalProvider
 from agent_engine.engine.langgraph.checkpointing import (
@@ -765,10 +771,9 @@ class LangGraphEngine(Engine):
         assert isinstance(node.node, OrchestratorSpec)
         spec = node.node
         path = node_id(node, parent_path)
-        primary_model = self._build_model(spec.model)
-        fallback_model = (
-            self._build_model(spec.model.fallback) if spec.model.fallback is not None else None
-        )
+        model = self._build_model(spec.model)
+        fb = spec.model.fallback
+        fallback_model = self._build_model(fb) if fb is not None else None
 
         children: list[ChildEntry] = []
         for child in node.children:
@@ -789,7 +794,7 @@ class LangGraphEngine(Engine):
         return OrchestratorNode(
             spec=spec,
             node_path=path,
-            model=primary_model,
+            model=model,
             children=children,
             filters=self._filters,
             base_dir=self._base_dir,
--- a/src/agent_engine/engine/langgraph/engine.py
+++ b/src/agent_engine/engine/langgraph/engine.py
@@ -765,7 +765,7 @@ class LangGraphEngine(Engine):
         assert isinstance(node.node, OrchestratorSpec)
         spec = node.node
         path = node_id(node, parent_path)
-        model = self._build_model(spec.model)
+        model = self._build_model_runnable(spec.model)
 
         children: list[ChildEntry] = []
         for child in node.children:
@@ -797,7 +797,7 @@ class LangGraphEngine(Engine):
         assert self._resolver_loader is not None
         assert self._hook_manager is not None
         tools, mcp_names, server_by_tool = self._build_agent_tools(spec)
-        bound_model = self._build_model(spec.model, tools=tools)
+        bound_model = self._build_model_runnable(spec.model, tools=tools)
         return AgentNode(
             spec=spec,
             node_path=node_path,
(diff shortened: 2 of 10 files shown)


AmitAvital1:
before i go deep review. have you checked it with real example? like examples/enterprise-knowledge-assistant
you can use ollama for free tokens :)

rishu685:
LGTM! Verified all automated tests pass (530/530). Also tested end-to-end on the enterprise-knowledge-assistant example using Ollama (llama3.2:3b) as the fallback model. The engine successfully catches primary model failures, executes the fallback model, and resumes execution seamlessly after tool approval interrupts. Looking forward for your feedback as well!

src/agent_engine/engine/langgraph/engine.py
```diff
@@ -814,13 +813,20 @@ def _build_agent_node(self, spec: AgentSpec, node_path: str) -> AgentNode:
...
+    def _build_model(
+        self, model: NodeModelConfig, tools: list[BaseTool] | None = None
+    ) -> BaseChatModel | Runnable:
+        primary_model = self._model_factory(
             model.provider,
             model.name,
             model.temperature,
             **_model_factory_kwargs(self._model_factory, model),
         )
+        bound_primary = primary_model.bind_tools(tools) if tools else primary_model
```
AmitAvital1, line 837: build model are responsibility to build only the model, no only binds tools this is make abuse to this function

src/agent_engine/engine/langgraph/engine.py
```diff
@@ -814,13 +813,20 @@ def _build_agent_node(self, spec: AgentSpec, node_path: str) -> AgentNode:
...
+        if model.fallback is not None:
+            bound_fallback = self._build_model(model.fallback, tools=tools)
```
AmitAvital1: I'm not lover of recursion, even if its simple from my experience its high risk for bugs and i don't see must have here

src/agent_engine/core/spec.py
```diff
@@ -14,6 +14,7 @@ class ModelConfig:
     region: str | None = None
     max_tokens: int | None = None
     top_p: float | None = None
+    fallback: ModelConfig | None = None
```
AmitAvital1: Ummm fallback that contain fallback? 
What about split ModelConfig to BaseModelConfig without fallback and now ModelConfig will holds embed BaseModelConfig and also fallback: BaseModelConfig | None = None?

AmitAvital1 (commented):
Claude input (need to verify)

Orchestrator + fallback model crashes at request time. engine.py:768 builds the orchestrator model via _build_model(spec.model) without tools. When a fallback is configured, _build_model returns a RunnableWithFallbacks, which is
  stored as OrchestratorNode._model. At request time nodes.py:614 does self._model.bind_tools(tools) — and RunnableWithFallbacks has no bind_tools. Since orchestrators always have child tools, any orchestrator with a fallback model    
  fails on every run.                                                                                                                                                                                                                      
                                                                                                                                                                                                                                           
  I reproduced it against the engine:                                                                                                                                                                                                      
  AttributeError: 'FakeRunnableWithFallbacks' object has no attribute 'bind_tools'                                                                                                                                                         
    src/agent_engine/engine/langgraph/nodes.py:614                                                                                                                                                                                         
  The real langchain RunnableWithFallbacks fails here too (its __getattr__ proxy blows up resolving bind_tools).

src/agent_engine/engine/langgraph/engine.py
```diff
@@ -814,13 +813,20 @@ def _build_agent_node(self, spec: AgentSpec, node_path: str) -> AgentNode:
             system_namespace=self._system_name,
         )
 
-    def _build_model(self, model: NodeModelConfig) -> BaseChatModel:
-        return self._model_factory(
+    def _build_model(
+        self, model: NodeModelConfig, tools: list[BaseTool] | None = None
+    ) -> BaseChatModel | Runnable:
```
AmitAvital1: BaseChatModel is not already runnable not?

src/agent_engine/engine/langgraph/engine.py
AmitAvital1, line 837: build model are responsibility to build only the model, no only binds tools this is make abuse to this function

src/agent_engine/engine/langgraph/engine.py
```diff
@@ -814,13 +813,20 @@ def _build_agent_node(self, spec: AgentSpec, node_path: str) -> AgentNode:
...
+        primary_model = self._model_factory(
             model.provider,
             model.name,
             model.temperature,
             **_model_factory_kwargs(self._model_factory, model),
         )
+        bound_primary = primary_model.bind_tools(tools) if tools else primary_model
+        if model.fallback is not None:
+            bound_fallback = self._build_model(model.fallback, tools=tools)
+            return bound_primary.with_fallbacks([bound_fallback], exceptions_to_handle=(Exception,))
```
AmitAvital1, line 841: this exception nessecery?

rishu685:
Thanks @AmitAvital1 for the detailed review! I've updated the PR addressing all suggestions:
 
 1. Model Building Responsibility & Recursion: Refactored `_build_model` so its sole responsibility is constructing the raw model instance without tool binding or recursion. Tool binding for static agent tools was extracted to `_build_model_runnable`.
 2. Structural Type Safety (`BaseModelConfig`): Split `ModelConfig` into `BaseModelConfig` (LLM parameters without a `fallback` field) and `ModelConfig(BaseModelConfig)` with `fallback: BaseModelConfig | None = None`. This structurally prevents nested fallbacks at the type level.
3. Orchestrator Fallback Handling: Fixed the request-time crash on orchestrators with fallback models. `OrchestratorNode` now receives raw primary and fallback models and dynamically binds child agent tools to both at request time. Added an automated test `test_orchestrator_fallback_model_execution` to verify this behavior.

src/agent_engine/engine/langgraph/nodes.py
```diff
@@ -522,10 +522,12 @@ def __init__(
         children: list[ChildEntry],
         filters: list[RouteFilter],
         base_dir: Path,
+        fallback_model: Any = None,
```
AmitAvital1: Any -> BaseChatModel ?

examples/enterprise-knowledge-assistant/mcps/local_knowledge_mcp/tools.py
```diff
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import hashlib
-
```
AmitAvital1, line 6: why have changes this file?

src/agent_engine/engine/langgraph/engine.py
```diff
@@ -765,7 +765,10 @@ def _build_orchestrator_node(
         assert isinstance(node.node, OrchestratorSpec)
         spec = node.node
         path = node_id(node, parent_path)
-        model = self._build_model(spec.model)
+        primary_model = self._build_model(spec.model)
+        fallback_model = (
```
AmitAvital1: fallback_model = self._build_model(spec.model.fallback) if spec.model.fallback is not None else None

AmitAvital1: no need ( ) i think

src/agent_engine/engine/langgraph/nodes.py
```diff
@@ -611,7 +613,18 @@ async def _run(
 
         # Build tools here so they share the live `visited` / `used_tools` lists.
         tools = [self._make_tool(e, state, visited, used_tools) for e in candidates]
-        bound_model = self._model.bind_tools(tools) if tools else self._model
+        if tools:
```
AmitAvital1: i think can do this if else better like

 bound_model = self._model.bind_tools(tools) if tools else self._model
        if self._fallback_model is not None:
            bound_fallback = (
                self._fallback_model.bind_tools(tools) if tools else self._fallback_model
            )
            bound_model = bound_model.with_fallbacks(
                [bound_fallback], exceptions_to_handle=(Exception,)
            )

verify it as well

AmitAvital1:
what about docs? we should add this to https://docs.extra-ai.co/docs/yaml-spec its new parameter

AmitAvital1:
docs should be focused thin and understanble

rishu685:
Done,made the suggested changes

AmitAvital1 (approved):

Issue #24, opened 2026-07-17
E2E test suite: example specs, generate correctness, mocked-LLM conversations


AmitAvital1:
## Background

#23 wires `make check` into CI as a required gate, but the test content it runs today doesn't verify the product end-to-end: that `generate` scaffolds the right files for a real spec, and that a served agent actually answers correctly. This issue adds that content — it's what #23's workflow ends up running.

Building blocks that already exist and should be reused, not reinvented: `examples/enterprise-knowledge-assistant/agents.yaml` (one full example spec), `CliRunner` for driving `agentctl` commands in tests (`tests/cli/test_serve_command.py`), and a `FakeChatModel` pattern for canned LLM answers — currently redefined separately in three different test files (`tests/engine/test_engine_flow.py`, `tests/runtime/test_engine_hooks.py`, `tests/runtime/test_tool_hooks.py`).

## Fixtures: both a full spec and small per-feature specs

One comprehensive spec alone means a failure doesn't say what broke; several tiny specs alone never test features interacting together. Use both:

- **Flagship** — keep `examples/enterprise-knowledge-assistant` as the one spec that exercises everything together (tools, resolvers, orchestrator routing, protected nodes, MCP).
- **Per-feature specs** — new, minimal, one feature each, so a failure points at what broke: e.g. `examples/minimal-agent` (single agent, no extras), `examples/with-resolver`, `examples/with-mcp`, `examples/protected-node`.

## Research: best e2e test tooling for Python

Don't assume `CliRunner` + FastAPI's in-process `TestClient` is automatically the right tool for driving a full generate → serve → converse flow — that's what today's unit tests use for narrower checks, not necessarily what an e2e suite should standardize on. Before implementing, evaluate options (in-process `TestClient`/`httpx.AsyncClient` vs. spawning the real `agentctl serve` process and hitting it over HTTP vs. a dedicated e2e framework) and note the tradeoff — mainly, in-process is faster and simpler but doesn't catch process-boundary/startup issues a real subprocess run would.

## Phase 1 — `generate` correctness

For every example spec: run `agentctl generate`, then assert every declared artifact exists on disk and matches what the spec declares — tool stubs, resolver stubs, `plugins/access.py`, MCP auth stubs, hook stubs (and prompt stubs, once #18 lands). Also assert a second `generate` run doesn't overwrite anything already filled in.

## Phase 2 — run + mocked-LLM conversations

[...]

PR #36, opened 2026-07-20
test(fixtures): add self-contained openai-default test system YAML

Karn2898:
## Summary
Add a minimal, validating agent system under `tests/fixtures/` for exercising the platform without the flagship example.

- `tests/fixtures/test_system.yaml`: root orchestrator routing to `greeting_agent` and `echo_agent`, with a shared resolver and an `echo_tool`.
- Defaults to the **openai** provider so it runs without an Anthropic key; any of `anthropic`/`openai`/`gemini`/`bedrock` is selectable via `provider:`.
- A comment in the YAML makes explicit that the matching API key is supplied via the environment (`OPENAI_API_KEY`, etc.), never in the YAML. `agentctl validate` needs no key; `agentctl run`/`serve` need a key for the chosen provider.
- Prompts, resolver stubs, and the `echo_tool` stub are implemented so `agentctl validate` passes fully offline.

## Verification
```bash
agentctl validate tests/fixtures/test_system.yaml
# => validation passed
```


AmitAvital1:
This is look greate example. but can you elaborate more how we running this test as system tests? how its should work? what makes you did it? (really helpfull question also for us to understand how to improve)

Thanks

Karn2898:
> This is look greate example. but can you elaborate more how we running this test as system tests? how its should work? what makes you did it? (really helpfull question also for us to understand how to improve)
> 
> Thanks

honestly this PR only adds the fixture , i doesn't include the test yet .
this fixture gives us everything a system test would need, but there is no pytest that actually runs the full flow against it. the good news is that the engine already supports this kind of system test.tests/engine/test_engine_flow.py already follows the pattern, it injects a fake model_factory into LangGraohEngine which returns a FakeChatModel. This fake model simulates a tool call and a final response so there is no API key or network dependency.

this fixture would be pretty straight forwrd: load  tests/fixtures/test_system.yaml and build the engine with fake model_factory, then send hello and verify it routes root router to greeting agent . send echo test and verify it routes root router to echo agent .
I left that test out of this PR intentionally. My goal here was to  validate the fixture first and the engine test also touches multiple internals, so I felt it deserved its own focused PR.If you are happy with the approach I can add it next.

AmitAvital1:
> > This is look greate example. but can you elaborate more how we running this test as system tests? how its should work? what makes you did it? (really helpfull question also for us to understand how to improve)
> > Thanks
> 
> honestly this PR only adds the fixture , i doesn't include the test yet . this fixture gives us everything a system test would need, but there is no pytest that actually runs the full flow against it. the good news is that the engine already supports this kind of system test.tests/engine/test_engine_flow.py already follows the pattern, it injects a fake model_factory into LangGraohEngine which returns a FakeChatModel. This fake model simulates a tool call and a final response so there is no API key or network dependency.
> 
> this fixture would be pretty straight forwrd: load tests/fixtures/test_system.yaml and build the engine with fake model_factory, then send hello and verify it routes root router to greeting agent . send echo test and verify it routes root router to echo agent . I left that test out of this PR intentionally. My goal here was to validate the fixture first and the engine test also touches multiple internals, so I felt it deserved its own focused PR.If you are happy with the approach I can add it next.

Very good plan! We really need this benefit. we already have issue https://github.com/extra-org/extra/issues/24 and this is great improvment.
How sound make this PR as infra for this issue? with couple infra test, and infra so any commit can be extended? and maybe next PR you can do more e2e tests cases

Karn2898:
> > > This is look greate example. but can you elaborate more how we running this test as system tests? how its should work? what makes you did it? (really helpfull question also for us to understand how to improve)
> > > Thanks
> > 
> > 
> > honestly this PR only adds the fixture , i doesn't include the test yet . this fixture gives us everything a system test would need, but there is no pytest that actually runs the full flow against it. the good news is that the engine already supports this kind of system test.tests/engine/test_engine_flow.py already follows the pattern, it injects a fake model_factory into LangGraohEngine which returns a FakeChatModel. This fake model simulates a tool call and a final response so there is no API key or network dependency.
> > this fixture would be pretty straight forwrd: load tests/fixtures/test_system.yaml and build the engine with fake model_factory, then send hello and verify it routes root router to greeting agent . send echo test and verify it routes root router to echo agent . I left that test out of this PR intentionally. My goal here was to validate the fixture first and the engine test also touches multiple internals, so I felt it deserved its own focused PR.If you are happy with the approach I can add it next.
> 
> Very good plan! We really need this benefit. we already have issue #24 and this is great improvment. How sound make this PR as infra for this issue? with couple infra test, and infra so any commit can be extended? and maybe next PR you can do more e2e tests cases

apologies for the late reply , Want me to draft the infra patch now ? I'm working on that  already btw

AmitAvital1:
> > > > This is look greate example. but can you elaborate more how we running this test as system tests? how its should work? what makes you did it? (really helpfull question also for us to understand how to improve)
> > > > Thanks
> > > 
> > > 
> > > honestly this PR only adds the fixture , i doesn't include the test yet . this fixture gives us everything a system test would need, but there is no pytest that actually runs the full flow against it. the good news is that the engine already supports this kind of system test.tests/engine/test_engine_flow.py already follows the pattern, it injects a fake model_factory into LangGraohEngine which returns a FakeChatModel. This fake model simulates a tool call and a final response so there is no API key or network dependency.
> > > this fixture would be pretty straight forwrd: load tests/fixtures/test_system.yaml and build the engine with fake model_factory, then send hello and verify it routes root router to greeting agent . send echo test and verify it routes root router to echo agent . I left that test out of this PR intentionally. My goal here was to validate the fixture first and the engine test also touches multiple internals, so I felt it deserved its own focused PR.If you are happy with the approach I can add it next.
> > 
> > 
> > Very good plan! We really need this benefit. we already have issue #24 and this is great improvment. How sound make this PR as infra for this issue? with couple infra test, and infra so any commit can be extended? and maybe next PR you can do more e2e tests cases
> 
> apologies for the late reply , Want me to draft the infra patch now ? I'm working on that already btw

Yea when ever you ready for review, push it inside this PR i would love to see it ! 🥇

Karn2898:
Hi I have pushed some files, hope it helps. Thank you

tests/fixtures/utils.py
```diff
@@ -0,0 +1,111 @@
+from __future__ import annotations
```
AmitAvital1, line 1: maybe lets call this file utils?

tests/fixtures/utils.py
```diff
@@ -0,0 +1,111 @@
...
+    name: str = "fake",
+    temperature: float | None = None,
+    *,
+    answer: str = "ok",
+) -> Any:
+    return FakeChatModel(answer=answer)
+
+
+FIXTURE_DIR = Path(__file__).resolve().parent
+FIXTURE_SPEC = FIXTURE_DIR / "test_system.yaml"
```
AmitAvital1: maybe lets do it injectable? think about we will added more yamls for tests in your infra

tests/fixtures/test_system/agents.yaml
```diff
@@ -0,0 +1,66 @@
...
+# it comes from the environment (e.g. OPENAI_API_KEY). `agentctl validate`
+# needs no key; `agentctl run`/`serve` need a key for the chosen provider.
+defaults:
+  model:
+    provider: openai
+    name: gpt-4o-mini
+    temperature: 0.0
+
+execution:
+  max_iterations: 10
```
AmitAvital1, line 12: what about test for all of this configurations?

AmitAvital1 (commented):
Nice work. can we also test generate?

AmitAvital1:
I see it that infra and support.py will be the infra, and test_yaml with generated files will be in dedicated folder that run this tests infra, and then we will add more folder with more yaml.. that extensabily more!

tests/fixtures/utils.py
Karn2898: sure

Karn2898:
> I see it that infra and support.py will be the infra, and test_yaml with generated files will be in dedicated folder that run this tests infra, and then we will add more folder with more yaml.. that extensabily more!

yep ,that's definetly a better structure , I have applied these changes

tests/fixtures/utils.py
Karn2898: That's  a good point . I have made changes ,just need to push

AmitAvital1:
> > I see it that infra and support.py will be the infra, and test_yaml with generated files will be in dedicated folder that run this tests infra, and then we will add more folder with more yaml.. that extensabily more!
> 
> yep ,that's definetly a better structure , I have applied these changes

i added more comments please see :)

tests/fixtures/test_system/agents.yaml
Karn2898: I added test_fixture_execution_policy_parsed_from_yaml which loads the spec from YAML and asserts all five fields (max_iterations, max_tool_calls, max_tool_calls_per_agent, max_child_agent_calls, allow_duplicate_tool_calls) match what's in the file. This guards against silent parser/schema changes dropping or misparsing those values. Not testing runtime enforcement here because tests/engine/test_execution_limits.py already covers that with inline specs; this fixture test just guarantees the YAML wiring is correct.

Karn2898:
> Nice work. can we also test generate?

Done , added tests/cli/test_generate_command.py with two tests using CliRunner against the fixture's agents.yaml

Karn2898:
> > > I see it that infra and support.py will be the infra, and test_yaml with generated files will be in dedicated folder that run this tests infra, and then we will add more folder with more yaml.. that extensabily more!
> > 
> > 
> > yep ,that's definetly a better structure , I have applied these changes
> 
> i added more comments please see :)

yeah I was ill so couldn't complete all tasks . now i will push

Asaf-prog:
Thanks — the reusable fixture and shared fake-model direction are valuable and align well with the system-test infrastructure described in #24.

I found one blocking issue in the echo-tool flow.

The shared `FakeChatModel` emits the tool call with:

```python
args={"message": ...}
```

but the fixture tool is defined as:

```python
def echo_tool(input: dict) -> str:
```

Since the engine builds a `StructuredTool` from the function signature, this does not match the expected input schema. The current test only verifies that an `echo_tool` record exists, so it can still pass when the tool execution actually failed.

Please align the tool signature and generated arguments, and explicitly assert that the tool record has `status == "succeeded"` and no error.

The PR is also currently not mergeable and needs to be rebased onto the latest `main`. While resolving that, please preserve the newer fake-model fallback behavior and fallback tests that now exist in `tests/engine/test_engine_flow.py`.

Once the branch is updated and the tool test proves successful execution rather than only attempted execution, the overall infrastructure looks useful.

Karn2898:
> Thanks — the reusable fixture and shared fake-model direction are valuable and align well with the system-test infrastructure described in #24.
> 
> I found one blocking issue in the echo-tool flow.
> 
> The shared `FakeChatModel` emits the tool call with:
> 
> ```python
> args={"message": ...}
> ```
> 
> but the fixture tool is defined as:
> 
> ```python
> def echo_tool(input: dict) -> str:
> ```
> 
> Since the engine builds a `StructuredTool` from the function signature, this does not match the expected input schema. The current test only verifies that an `echo_tool` record exists, so it can still pass when the tool execution actually failed.
> 
> Please align the tool signature and generated arguments, and explicitly assert that the tool record has `status == "succeeded"` and no error.
> 
> The PR is also currently not mergeable and needs to be rebased onto the latest `main`. While resolving that, please preserve the newer fake-model fallback behavior and fallback tests that now exist in `tests/engine/test_engine_flow.py`.
> 
> Once the branch is updated and the tool test proves successful execution rather than only attempted execution, the overall infrastructure looks useful.

echo_tool signature now matches the emitted args (message: str instead of input: dict)
Test now asserts status == "succeeded" and error is None
Branch rebased onto latest main; fallback behavior/tests in test_engine_flow.py preserved

Asaf-prog:
Please rebase this PR onto `main`.

Karn2898:
> Please rebase this PR onto `main`.

Rebased onto the latest main and force-pushed the updated branch. I also resolved the merge conflicts while preserving the newer fake-model fallback coverage and the CLI generate test additions.

Asaf-prog (approved):

Karn2898:
Thank you sir :)

PR #39, opened 2026-07-21
fix(orchestrator): load both system and orchestrator prompt files

### LINKED ISSUE #38: Orchestrator prompt not attached: falls back to description, losing context
## Current behavior

`OrchestratorNode.__call__` builds the orchestrator's prompt from the wrong field, and only uses one of the two prompt files:

https://github.com/extra-org/extra/blob/main/src/agent_engine/engine/langgraph/nodes.py#L535

```python
base_prompt = load_file(self._base_dir, self._spec.prompts.system) or self._spec.description
```

Per the documented contract (`docs/PROMPT_RENDERING.md`), an orchestrator node has **two distinct, complementary** prompt files:

| Field | Purpose | Required |
| --- | --- | --- |
| `system` | node behavior/persona | no |
| `orchestrator` | routing instructions | yes |

Both are meant to be attached to the LLM call. Instead, the code only ever reads `.system` and falls back to the one-line YAML `description` when it's empty — `.orchestrator` (the required routing-instructions file) is never read at all, and `.system` and `.orchestrator` are never combined.

Since most orchestrator configs (see `examples/enterprise-knowledge-assistant/agents.yaml`) set both `system` and `orchestrator` files, in practice: `system.md` content silently replaces `orchestrator.md` content, and the routing instructions the orchestrator actually needs are dropped entirely, with no error or warning.

For contrast, `AgentNode` (`nodes.py:187`) correctly uses `self._spec.prompts.system`, since agents only have one prompt field (`system`). Looks like this orchestrator branch was copy-pasted from the agent branch without accounting for the orchestrator's two-prompt shape.

## Steps to reproduce

1. Define an orchestrator node in `agent.yml` with a `description`, a `prompts.system` file, and a `prompts.orchestrator` file, e.g.:
   ```yaml
   orchestrators:
     research_router:
       description: "Routes research queries"
       prompts:
         system: prompts/research_router/system.md
         orchestrator: prompts/research_router/orchestrator.md
   ```
2. Put distinct, identifiable content in each file (e.g. persona text in `system.md`, routing rules in `orchestrator.md`).
3. Run the graph and inspect the system prompt actually sent to the orchestrator's LLM call (or log `base_prompt` in `OrchestratorNode.__call__`).
4. Observe only `system.md`'s content appears (or `description`, if `system` isn't set) — `orchestrator.md`'s routing instructions never appear anywhere in the prompt.

## Goal

`OrchestratorNode.__call__` should load **both** `self._spec.prompts.system` and `self._spec.prompts.orchestrator` and combine them into the final prompt (plus `_ORCHESTRATOR_CONTRACT`), not read one instead of the other. `description` should remain only as a last-resort fallback when neither prompt file is configured.

Oberonskii:
## Description

`OrchestratorNode.__call__` was only reading `self._spec.prompts.system`, silently dropping the orchestrator prompt file (`self._spec.prompts.orchestrator`). This meant routing instructions configured in `orchestrator.md` were never sent to the LLM, causing the orchestrator to lose its routing context and fall back to the brief YAML `description` field.

### What was changed

In `src/agent_engine/engine/langgraph/nodes.py`, the `OrchestratorNode.__call__` method now loads **both** `system` and `orchestrator` prompt files and combines them into the final prompt. When neither file is configured, `description` is used as a last-resort fallback, preserving backward compatibility.

### How it was tested

- All 525 existing tests pass (including 14 engine flow tests that exercise orchestrator routing)
- The change follows the same `load_file` pattern already used by `AgentNode._build_prompt`
- Backward compatible: when both prompt files are absent (as in existing tests with empty `OrchestratorPromptSet()`), behavior falls through to `description` unchanged

### Related issue

- Closes #38


uv.lock
```diff
@@ -0,0 +1,3059 @@
+version = 1
```
AmitAvital1: We are working with the build tool that call poertry please make an alignment

AmitAvital1 (changes requested):
We are working with the build tool that call poertry please make an alignment

AmitAvital1:
Please star the project so i can assign to you the issue

Oberonskii:
**Update:** Habe die `uv.lock` entfernt. Das Projekt verwendet Poetry fürs Dependency-Management – das war mein Fehler, die `uv.lock` hatte da nichts zu suchen. Der eigentliche Bugfix (beide Prompt-Files laden) bleibt erhalten. Nur noch 5 Zeilen Diff. 👍

Oberonskii:
Removed the `uv.lock` file — this project uses Poetry for dependency management, so the lockfile was out of place. The actual bugfix (loading both prompt files) is unchanged. Down to just 5 lines of diff now.

Asaf-prog:
please write in english.
in addition, can you share a screenshot of lungfuse? it a critical bug and we need an extra validation before merge.
you need to create an api key in langfuse and add it inside the .env file, and the engine will automatically integrate and present the stack

AmitAvital1:
Done by other

PR #40, opened 2026-07-22
fix: publish multi-arch container image (linux/amd64 + linux/arm64) — closes #37

### LINKED ISSUE #37: Published container image is amd64-only — arm64 users pay a QEMU emulation tax on every pull
## Background

The release workflow (`.github/workflows/release-please.yml`) builds and pushes the container image with `docker/build-push-action@v6` and no `platforms` set — on `ubuntu-latest`, that produces a single `linux/amd64` image at `ghcr.io/extra-org/extra`. There's no QEMU or buildx setup in the workflow itself.

Anyone pulling that image on Apple Silicon (or another arm64 host) gets it emulated locally by Docker Desktop's own QEMU layer — slower startup, slower runtime, every single pull, on every developer's machine.

## Purpose

Publish a multi-arch image (`linux/amd64` + `linux/arm64`) under the same tag, so an arm64 host pulls its native build directly instead of emulating the amd64 one. The emulation cost moves from every developer's laptop, on every pull, to the CI runner, once per release.

## Worth knowing going in

- Cross-building `arm64` on the (amd64) GitHub runner still uses QEMU — just once, in CI, at release time, instead of on every developer's machine on every pull.
- Multi-arch builds take noticeably longer than single-arch ones, since the foreign-arch build is itself emulated during the build step. Worth watching the actual added CI time once this lands, since `publish-container` only runs on a real release, not every push.
- Only applies going forward — existing published tags stay single-arch amd64 until the next release is cut under the new workflow.

rishu685:
Problem
The publish-container job in .github/workflows/release-please.yml builds
and pushes a single linux/amd64 image with no platforms set and no QEMU or
Buildx setup. Developers on Apple Silicon (arm64) and other ARM hosts are forced
to emulate the amd64 image via Docker Desktop's QEMU layer on every pull —
slower startup, slower runtime, on every machine, every time.

Solution
Configure the release workflow to build and publish a true multi-arch manifest
(linux/amd64 + linux/arm64) under the same tag. An arm64 host now pulls its
native build directly. The QEMU emulation cost is moved from every developer's
machine on every pull to the CI runner, once per release.

Changes
.github/workflows/release-please.yml
Add docker/setup-qemu-action@v3 — enables QEMU for cross-arch builds on the
amd64 GitHub runner.
Add docker/setup-buildx-action@v3 — switches to the docker-container
BuildKit driver required for multi-platform manifest pushes.
Set platforms: linux/amd64,linux/arm64 on docker/build-push-action@v6.
Set provenance: false — suppresses SLSA attestation manifests so the GHCR
manifest list stays clean (just the two arch entries).
Add cache-from: type=gha + cache-to: type=gha,mode=max — caches all
intermediate BuildKit layers in GitHub Actions, keeping subsequent release
build times reasonable despite QEMU-emulated cross-compilation.

Closes #37

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,10 +2,8 @@ FROM python:3.11-slim
 
 WORKDIR /app
 COPY pyproject.toml README.md ./
-RUN pip install --no-cache-dir .
-
 COPY src/ ./src/
-RUN pip install --no-cache-dir --no-deps .
+RUN pip install --no-cache-dir .
 
 COPY entrypoint.sh /entrypoint.sh
 RUN chmod +x /entrypoint.sh
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,8 +2,10 @@ FROM python:3.11-slim
 
 WORKDIR /app
 COPY pyproject.toml README.md ./
-COPY src/ ./src/
 RUN pip install --no-cache-dir .
+
+COPY src/ ./src/
+RUN pip install --no-cache-dir --no-deps .
 
 COPY entrypoint.sh /entrypoint.sh
 RUN chmod +x /entrypoint.sh
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -32,6 +32,12 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
       - uses: docker/login-action@v3
         with:
           registry: ghcr.io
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -2,9 +2,11 @@ FROM python:3.11-slim
 
 WORKDIR /app
 COPY pyproject.toml README.md ./
-COPY src/ ./src/
 RUN pip install --no-cache-dir .
 
+COPY src/ ./src/
+RUN pip install --no-cache-dir --no-deps .
```
AmitAvital1: also in line 5 pip install?

Dockerfile
AmitAvital1, line 1: why we need change at this file?

AmitAvital1 (approved):

Issue #31, opened 2026-07-18
Support prompt caching to reduce cost on repeated system prompts


AmitAvital1:
## Background

Every request to an agent resends its full system prompt to the model provider, and that system prompt is frequently the same or nearly the same text across many requests — it's read from a fixed file per node. Several model providers offer prompt caching: a way to mark reused content so the provider charges less (and responds faster) on subsequent requests that repeat it, instead of billing full price for the same tokens every time.

## Purpose

Let a node opt into provider-side prompt caching for its system prompt, so repeated requests to the same agent don't pay full input-token cost for content that hasn't changed.

## Proposed YAML

```yaml
model:
  provider: anthropic
  name: claude-sonnet-4-6
  cache_system_prompt: true   # default true
```

Nests next to `model`'s other fields (same place #25's `fallback` field lives), usable wherever a model is defined — per-agent, per-orchestrator, or `defaults.model`. Default **on**: most agents send the same system prompt repeatedly, so caching should be the default, not something every spec has to opt into.

## Research

- **Cross-provider, not just Anthropic.** Extra supports Anthropic, Bedrock, Gemini, and OpenAI-compatible providers (`src/agent_engine/models/factory.py`) — caching mechanics differ across all of them, not just in whether it's supported. Some cache automatically server-side with no flag at all; some need explicit opt-in markup; some charge a premium on the write and only save on a hit, which matters for low-traffic agents. Needs a provider-by-provider pass, not an Anthropic-shaped assumption applied everywhere.
- **Only `system` (and `orchestrator`) prompts apply.** `prompts.user` exists in the YAML schema but isn't read anywhere by the engine at runtime today — the live user message always comes from the actual conversation, never from a file. There's currently nothing to cache there.
- **Caching part of one message is possible, but the prompt isn't built that way today.** Several providers cache at a content-block level — a message's content can be a list of blocks, with the cache boundary marked on one of them, so a static prefix can be cached separately from dynamic content that follows it. Today's system prompt is built by interpolating resolver values inline into the template text (`render_prompt()`), producing one merged string rather than separate static/dynamic blocks. Achieving prefix caching means restructuring how the message is assembled, not just adding a client-side flag.

PR #41, opened 2026-07-22
fix[models]: support cache_system_prompt parameter for model configu…

rishu685:
This PR addresses #31 by introducing a new `cache_system_prompt` field to the model configuration (`ModelConfig`). This field defaults to `True` and controls whether the system prompt and tool definitions are cached on supported providers (specifically Anthropic).

## Scope & Implementation Details
1. Spec & Dataclass: Added `cache_system_prompt: bool = True` to the `BaseModelConfig` dataclass in `src/agent_engine/core/spec.py`.
2. YAML Validation & Parser:
   - Updated `_validate_model` in `src/agent_engine/parsers/yaml/parser.py` to ensure `cache_system_prompt` is validated as a boolean.
   - Updated `_build_model` and `_build_base_model` to correctly parse `cache_system_prompt` from the YAML definitions, defaulting to `True`.
3. Model Factory:
   - Added `cache_system_prompt` parameter to `build_chat_model` in `src/agent_engine/models/factory.py`.
   - Updated `_build_anthropic_model` to conditionally pass `cache_control={"type": "ephemeral"}` parameter via `init_chat_model()` kwargs when `cache_system_prompt` is `True`.
4. Engine Configuration:
   - Updated `_model_factory_kwargs` in `src/agent_engine/engine/langgraph/engine.py` to forward `cache_system_prompt` from node configurations to the model factory.
5. Formatting & Quality Fixes:
   - Fixed some pre-existing long lines (`E501`) and sorted imports in `nodes.py` and `engine.py`.

## Verification
- Added YAML configuration parser tests under `tests/parsers/test_cache_system_prompt.py`.
- Added model factory test coverage under `tests/models/test_factory.py`.
- Verified that the full test suite (`make check`) passes cleanly:
  - `ruff check src tests`: Passing
  - `mypy src tests`: Passing
  - `agentctl validate examples/enterprise-knowledge-assistant/agents.yaml`: Passing
  - `pytest`: Passing (541 tests passed)

### DIFF
--- a/src/agent_engine/core/spec.py
+++ b/src/agent_engine/core/spec.py
@@ -14,6 +14,7 @@ class ModelConfig:
     region: str | None = None
     max_tokens: int | None = None
     top_p: float | None = None
+    cache_system_prompt: bool = True
 
 
 @dataclass(frozen=True)
--- a/src/agent_engine/engine/langgraph/engine.py
+++ b/src/agent_engine/engine/langgraph/engine.py
@@ -169,6 +169,7 @@ def _model_factory_kwargs(factory: ModelFactory, model: NodeModelConfig) -> dict
         "region": model.region,
         "max_tokens": model.max_tokens,
         "top_p": model.top_p,
+        "cache_system_prompt": model.cache_system_prompt,
     }
     present: dict[str, object] = {
         key: value for key, value in optional.items() if value is not None
--- a/src/agent_engine/models/factory.py
+++ b/src/agent_engine/models/factory.py
@@ -34,6 +34,7 @@ def build_chat_model(
     region: str | None = None,
     max_tokens: int | None = None,
     top_p: float | None = None,
+    cache_system_prompt: bool = True,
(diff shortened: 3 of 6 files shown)


AmitAvital1:
I understand this commit is specifically for anthropic, what about other provider? we will do on other commits? or you think its harder?
what about enforce it only for anthropic?

rishu685:
See what i think is prompt caching is handled differently across providers:

--Anthropic is the only one requiring this explicit client-side flag (cache_control) here.
--OpenAI caches automatically behind the scenes (for >1024 tokens) without any parameter so i don't think we should enforce it here.
--Gemini and Bedrock don't support simple pass-through caching flags at initialization.
Because of this, the parameter is ignored for other providers. We chose not to fail validation for non-Anthropic models so developers don't have to keep adding or removing this line when switching providers in their YAML.

Let me know if you'd prefer us to strictly enforce this flag only for Anthropic, or make the change for others!

PR #52, opened 2026-07-26
fix(orchestrator): load both system and orchestrator prompt files

### LINKED ISSUE #38: Orchestrator prompt not attached: falls back to description, losing context

rishu685:
`OrchestratorNode.__call__` was previously only reading `self._spec.prompts.system`, silently dropping the orchestrator prompt file (`self._spec.prompts.orchestrator`). This meant routing instructions configured in `orchestrator.md` were never sent to the LLM, causing the orchestrator to lose its routing context.

### What was changed
1. Updated `OrchestratorNode.__call__` in `src/agent_engine/engine/langgraph/nodes.py` to load*both `system` and `orchestrator` prompt files and combine them.
2. Kept `description` as a last-resort fallback when neither prompt file is configured, preserving backward compatibility.
3. Added a comprehensive async test case `test_orchestrator_loads_both_prompts` in `tests/engine/test_engine_flow.py` covering all four possible configurations:
   - Both system and orchestrator prompts provided.
   - Only system prompt provided.
   - Only orchestrator prompt provided.
   - Neither provided (falls back to description).
4. Cast the mock models to `Any` using `cast(Any, ...)` to satisfy strict type-checking checks.

### Verification Results
- **Pytest**: `533 passed` (including the new test cases).
- **Ruff**: `All checks passed!`
- **Mypy**: `Success: no issues found in 167 source files`.

### Related Issue
- Closes #38

### DIFF
--- a/src/agent_engine/engine/langgraph/nodes.py
+++ b/src/agent_engine/engine/langgraph/nodes.py
@@ -535,7 +535,17 @@ class OrchestratorNode:
 
     async def __call__(self, state: GraphState) -> dict[str, object]:
         candidates = self._filter_children(state)
-        base_prompt = load_file(self._base_dir, self._spec.prompts.system) or self._spec.description
+        # orchestrator is the required routing prompt; system is an optional persona prefix.
+        orchestrator_content = load_file(self._base_dir, self._spec.prompts.orchestrator)
+        if orchestrator_content:
+            system_content = load_file(self._base_dir, self._spec.prompts.system)
+            base_prompt = (
+                f"{system_content}\n\n{orchestrator_content}"
+                if system_content
+                else orchestrator_content
+            )
+        else:
+            base_prompt = self._spec.description
         system_prompt = f"{base_prompt}\n{_ORCHESTRATOR_CONTRACT}"
         return await self._run(system_prompt, candidates, state)
 
--- a/src/agent_engine/engine/langgraph/nodes.py
+++ b/src/agent_engine/engine/langgraph/nodes.py
@@ -535,17 +535,9 @@ class OrchestratorNode:
 
     async def __call__(self, state: GraphState) -> dict[str, object]:
         candidates = self._filter_children(state)
-        # orchestrator is the required routing prompt; system is an optional persona prefix.
+        base_prompt = load_file(self._base_dir, self._spec.prompts.system) or self._spec.description
         orchestrator_content = load_file(self._base_dir, self._spec.prompts.orchestrator)
-        if orchestrator_content:
-            system_content = load_file(self._base_dir, self._spec.prompts.system)
-            base_prompt = (
-                f"{system_content}\n\n{orchestrator_content}"
-                if system_content
-                else orchestrator_content
-            )
-        else:
-            base_prompt = self._spec.description
+        base_prompt = f"{base_prompt}\n\n{orchestrator_content}"
         system_prompt = f"{base_prompt}\n{_ORCHESTRATOR_CONTRACT}"
         return await self._run(system_prompt, candidates, state)
 
--- a/tests/engine/test_engine_flow.py
+++ b/tests/engine/test_engine_flow.py
@@ -589,8 +589,7 @@ async def test_orchestrator_loads_both_prompts(tmp_path: Path) -> None:
     )
     assert model_both.captured_prompt == expected_both
 
-    # 2. Test system prompt only (no orchestrator) — falls back to description because
-    # orchestrator is required; system alone is not enough to build the prompt.
+    # 2. Test system prompt only (no orchestrator) — uses system as the base prompt.
     spec_sys = OrchestratorSpec(
         id="orch",
         name="orch",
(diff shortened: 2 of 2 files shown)


src/agent_engine/engine/langgraph/nodes.py
```diff
@@ -535,7 +535,10 @@ def __init__(
 
     async def __call__(self, state: GraphState) -> dict[str, object]:
         candidates = self._filter_children(state)
-        base_prompt = load_file(self._base_dir, self._spec.prompts.system) or self._spec.description
+        system_content = load_file(self._base_dir, self._spec.prompts.system)
+        orchestrator_content = load_file(self._base_dir, self._spec.prompts.orchestrator)
+        parts = [p for p in (system_content, orchestrator_content) if p]
```
AmitAvital1: orchestrator prompt is required field, and can do this check if system_content only for system contnet which not required.
This method can make reader mistaken so think orchestrator prompt are not required.

AmitAvital1 (changes requested):
small comment and can be approved
nice fast work 🚀

rishu685:
Done made the required changes!

src/agent_engine/engine/langgraph/nodes.py
```diff
@@ -535,7 +535,17 @@ def __init__(
 
     async def __call__(self, state: GraphState) -> dict[str, object]:
         candidates = self._filter_children(state)
-        base_prompt = load_file(self._base_dir, self._spec.prompts.system) or self._spec.description
+        # orchestrator is the required routing prompt; system is an optional persona prefix.
```
AmitAvital1: comment not needed

src/agent_engine/engine/langgraph/nodes.py
```diff
@@ -535,7 +535,17 @@ def __init__(
...
+        orchestrator_content = load_file(self._base_dir, self._spec.prompts.orchestrator)
+        if orchestrator_content:
```
AmitAvital1: why if?

src/agent_engine/engine/langgraph/nodes.py
```diff
@@ -535,7 +535,13 @@ def __init__(
 
     async def __call__(self, state: GraphState) -> dict[str, object]:
         candidates = self._filter_children(state)
-        base_prompt = load_file(self._base_dir, self._spec.prompts.system) or self._spec.description
+        system_content = load_file(self._base_dir, self._spec.prompts.system)
```
AmitAvital1: Hi sorry maybe my comments are not clear enough. here you do changing behavior more than what need.
We should do as was before
base prompt =  base_prompt = load_file(self._base_dir, self._spec.prompts.system) or self._spec.description

then you should join the orchestrator prompt that have to be exist as schema.
no if else and for loop.

Am i missing something?

rishu685: done with the suggested changes can you please review if the logic is correct now?

rishu685: Just wanted to flag one thing — in the schema, orchestrator is actually defined as orchestrator: str = "" (defaults to the empty string), so it's technically optional and not always guaranteed to be set.

Because of that removing the if check could result in an empty line gap in the prompt when orchestrator isn't configured. The if orchestrator_content: guard prevents that.

Happy to remove this if you're okay with that edge case, since in practice every orchestrator node should always have a routing prompt set. Just wanted to make sure we're aligned before changing it!

AmitAvital1: No need the if orchestrator content 
Orchestrator prompt should be required, and check should be done on validation. If not we will open an issue :) 
So please remove the if orchestrator and then we are good to go

rishu685: done!

AmitAvital1 (approved):

PR #53, opened 2026-07-26
fix: show clear message when widget hits token budget (429)

### LINKED ISSUE #50: Widget shows generic "Something went wrong" when a conversation hits the token budget (429)
## Summary

When a conversation reaches its `context_max_tokens` budget, the backend correctly returns **HTTP 429** (`ConversationTokenBudgetExceeded`), but the chat widget renders the generic error **"Something went wrong. Please try again."**

This is misleading: it implies a transient failure and tells the user to retry, but every retry will 429 again — the conversation is permanently over budget. The user is given no indication that they hit a context limit or that they need to start a new conversation.

## Steps to reproduce

1. Run the manager with a budget set, e.g. `CONTEXT_MAX_TOKENS=2000` (see `examples/starter/.env.example`).
2. Open the widget and chat until cumulative tokens exceed the budget (the context meter reaches ~100%).
3. Send one more message.

## Actual behavior

The assistant bubble shows **"Something went wrong. Please try again."**

Server logs (confirming the 429):
```
manager-1  | INFO: POST /conversations/<id>/messages/stream HTTP/1.1  429 Too Many Requests
manager-1  | INFO: POST /conversations/<id>/messages        HTTP/1.1  429 Too Many Requests
manager-1  | INFO: GET  /conversations/<id>/usage           HTTP/1.1  200 OK
```
(The second 429 is the non-streaming fallback; the 200 is the post-turn usage refresh.)

## Expected behavior

A clear, non-retry message, e.g. **"This conversation has reached its context limit. Start a new chat to continue."** — ideally without the "Please try again" call to action, since retrying cannot succeed.

## Root cause

The widget only special-cases **404** (stale conversation recovery). A **429** is treated like any other failure:

- `src/agent_manager/api/static/widget/react/useConversation.ts` — `isMissingConversation()` matches only `status === 404`; `send`/`stream` re-throw on 429.
- `src/agent_manager/api/static/widget/react/AgentChatApp.tsx` — `submit()` catches the stream error and calls `sendWithoutStreaming()`, whose `catch` replaces the entry with the constant `GENERIC_ERROR`, regardless of status.
- Backend raises the 429 in both `send_message` and `stream_message` in `src/agent_manager/api/routes.py` (`ConversationTokenBudgetExceeded` -> 429).

## Suggested fix

Detect `AgentChatHttpError.status === 429` in the submit path and render a distinct, budget-specific message (reusing the existing `.msg-error` alert styling). Skip the non-streaming fallback for 429, since it will also 429. Optionally surface a "New chat" affordance to reset the stored conversation id.

VedantPatel04:
Skip the non-streaming fallback on ConversationTokenBudgetExceeded and tell users to start a new chat instead of the "Something went wrong" message displayying.

Tested using `make check` as well, run is clean.

Closes #50

### DIFF
--- a/src/agent_manager/api/static/widget.js
+++ b/src/agent_manager/api/static/widget.js
@@ -52922,6 +52922,7 @@ function statusLabel(state) {
 
 // src/agent_manager/api/static/widget/react/AgentChatApp.tsx
 var import_jsx_runtime4 = __toESM(require_jsx_runtime(), 1);
+var TOKEN_BUDGET_EXCEEDED_MESSAGE = "This conversation has reached its context limit. Start a new chat to continue.";
 var nextMessageCounter = 0;
 function nextMessageId(role) {
   nextMessageCounter += 1;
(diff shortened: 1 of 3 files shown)


AmitAvital1:
Hi, FE layer should be stateless as much as can, who handling the error should be BE and FE only parsing it or maybe detailed it with human read able string. GENERIC ERROR should be only if we got error not known but here we should upstream dedicated error

AmitAvital1 (changes requested):
BTW As i said some other already opened PR so you can work in parallel but you can just pick other issue :)

AmitAvital1:
Are you going to finalize it?

VedantPatel04:
> Are you going to finalize it?

Hi sorry, I'm away from my computer for a few days so I'll step back on this issue and let the other contributor's PR through. I'll be on the lookout for any future such issues. 

Thanks for updating me Amit :).

PR #74, opened 2026-08-01
fix(generate): report unreferenced tools

### LINKED ISSUE #67: generate: silently omits tools declared but not referenced by any agent
## Bug

If you declare a tool at the top level but never add it to any agent's `tools` list, `agentctl generate` says nothing about it. It doesn't create a stub (expected, since nothing uses it), but it also doesn't appear anywhere in the output — not in `create`, not in `skip`. It's just invisible, so a forgotten reference looks identical to an intentional one.

```yaml
tools:
  add_new_user:
    description: Adding new user to the system

agents:
  admin_agent:
    prompts:
      system: "prompts/scoping_agent/system.md"
    tools: []   # add_new_user never referenced here
```

## Steps to reproduce

1. Save the YAML above as `agents.yml`.
2. Run:

   ```bash
   agentctl generate --config agents.yml
   ```

## Actual

`add_new_user` doesn't show up in the output at all. Output only says "Nothing to generate" or lists unrelated stubs — no trace that a declared tool was skipped.

## Expected

`generate` should report it explicitly, distinct from `create`/`skip`:

```
  ignore  add_new_user  (declared but not referenced by any agent)
```

Still no stub file — that's correct, nothing calls it. But the tool should show up in the output instead of disappearing.

## Where to fix

`src/agent_engine/generate/generator.py` — `Generator.generate()` collects `tool_ids` from the graph via `_collect()` (line 216), but never looks at the top-level `tools:` registry to see what was declared and skipped. Need to diff the registry against `tool_ids` and report the difference in `GenerateResult` (or a new field alongside `created`/`skipped`), surfaced by `src/agentctl/main.py`'s `generate` command.

happykawayigt:
## Summary

- preserve the complete top-level tool registry in the compiled system spec
- report tools that are declared but not referenced by any agent as `ignore`
- keep ignored tools out of generated stubs and `plugins.toml`
- cover direct generator behavior and first/idempotent CLI runs

Fixes #67

## Validation

- targeted regression tests: 2 passed
- full pytest excluding five existing Windows path-separator assertions in `tests/generate/test_plugins_manifest.py`: 555 passed
- `mypy src tests`: 169 source files passed
- Ruff check and format check for all changed files
- `git diff --check`

## AI assistance

Implemented and reviewed with OpenAI Codex; all listed validation commands were run locally.

### DIFF
--- a/src/agent_engine/core/spec.py
+++ b/src/agent_engine/core/spec.py
@@ -168,3 +168,4 @@ class SystemSpec:
     hooks: HooksConfig = field(default_factory=HooksConfig)
     plugins: PluginsConfig = field(default_factory=PluginsConfig)
     execution: ExecutionPolicy = field(default_factory=ExecutionPolicy)
+    tools: tuple[ToolSpec, ...] = field(default_factory=tuple)
--- a/src/agent_engine/generate/generator.py
+++ b/src/agent_engine/generate/generator.py
@@ -15,6 +15,7 @@ from agent_engine.generate.manifest import (
 class GenerateResult:
     created: list[str] = field(default_factory=list)
     skipped: list[str] = field(default_factory=list)
+    ignored: list[str] = field(default_factory=list)
 
 
 class Generator:
@@ -37,6 +38,7 @@ class Generator:
             has_protected,
             mcp_plugin_ids,
         ) = _collect(spec.graph)
+        result.ignored.extend(tool.id for tool in spec.tools if tool.id not in tool_ids)
 
         tools_dir = base_dir / "plugins" / "tools"
         tools_dir.mkdir(parents=True, exist_ok=True)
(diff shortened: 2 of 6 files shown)


AmitAvital1:
Hi @happykawayigt please see discord message and follow this instructions

Asaf-prog:
Thanks, this looks good and matches the intended behavior from #67.

The implementation keeps the complete top-level tool registry in `SystemSpec`, reports tools that are declared but not referenced by the compiled agent graph, and correctly avoids generating stubs or manifest entries for them.

The CLI output is clear, and I like that the tests cover both the initial generation and the idempotent second run.

One small non-blocking suggestion: since the PR also guarantees that ignored tools stay out of `plugins.toml`, it would be useful to add an explicit assertion for that in the generator test.

Otherwise, the scope and implementation look good to me. Once it is ready, please mark the PR as ready for review.

Asaf-prog (approved):

PR #75, opened 2026-08-01
fix(api): sanitize raw tool exception text to generic error in ToolRecord

### LINKED ISSUE #69: api: raw tool exception text is sent to the client instead of a generic message
## Bug

When a tool raises an exception, the raw exception text is sent all the way to the chat UI and rendered verbatim to the end user — internal hostnames, ports, and API paths included.

Example seen in the widget:

```
add_new_user · local
Error
HTTPConnectionPool(host='localhost', port=3000): Max retries exceeded with url: /api/v1/auths/add ...
```

## Steps to reproduce

1. Have a tool implementation that raises an exception exposing internal details (e.g. a failed HTTP call to an internal service).
2. Trigger the tool from a chat message.
3. Look at the tool card in the widget, or the raw API response.

## Actual

The exact `str(exc)` (truncated to 200 chars) reaches the client unfiltered, through two paths:

- `POST /conversations/{id}/messages` → `InvokeResponse.used_tools` ([app.py:258](https://github.com/extra-org/extra/blob/main/src/agent_engine/api/app.py#L258))
- `POST /conversations/{id}/messages/stream` → the `final` SSE event's `used_tools` ([app.py:337](https://github.com/extra-org/extra/blob/main/src/agent_engine/api/app.py#L337))

Both build `ToolRecord(**dataclasses.asdict(t))` directly from the internal `ToolUsageRecord`, which carries `error = str(exc)[:200]` set in `_record_error` ([nodes.py:351](https://github.com/extra-org/extra/blob/main/src/agent_engine/engine/langgraph/nodes.py#L351)). Nothing sanitizes it before it's serialized to JSON. The widget then renders it as-is with no filtering (`ToolOutput`, [shadcnAiElements.tsx:190-192](https://github.com/extra-org/extra/blob/main/src/agent_manager/api/static/widget/react/shadcnAiElements.tsx#L190)).

This is separate from the existing generic-error handling for HTTP 500s ([AgentChatApp.tsx:199](https://github.com/extra-org/extra/blob/main/src/agent_manager/api/static/widget/react/AgentChatApp.tsx#L199)) — a failed tool call is not an HTTP error, it's a "successful" response that happens to contain a failed tool record, so that existing guard never applies here.

## Expected

The end user sees only that the tool failed, e.g.:

```
add_new_user · local
Error
Tool execution failed
```

The real error stays available to:
- **Server logs** — already captured today via the `on_tool_error` hook → logging observability provider (`tool end status=error error=...`, [logging/provider.py:40](https://github.com/extra-org/extra/blob/main/src/agent_engine/observability/providers/logging/provider.py#L40)).
- **The model itself** — unchanged. `_record_error` returns `f"Tool error: {exc}"` as the tool's result text so the LLM can read the real error and attempt recovery ([nodes.py:360](https://github.com/extra-org/extra/blob/main/src/agent_engine/engine/langgraph/nodes.py#L360)). This is intentional and should stay as-is — only the human-facing UI copy needs to change.

## Proposed solution

Sanitize at the API boundary, not deeper in the engine — `ToolUsageRecord.error` should keep carrying the real exception (logs, hooks, and the model all depend on it as-is). Replace it with a generic string only when building the client-facing `ToolRecord` in `app.py`, at both call sites (`/messages` and `/messages/stream`):

```python
def _client_tool_record(t: ToolUsageRecord) -> ToolRecord:
    fields = dataclasses.asdict(t)
    if fields.get("error"):
        fields["error"] = "Tool execution failed"
    return ToolRecord(**fields)
```

Used in place of `ToolRecord(**dataclasses.asdict(t))` at [app.py:258](https://github.com/extra-org/extra/blob/main/src/agent_engine/api/app.py#L258) and [app.py:337](https://github.com/extra-org/extra/blob/main/src/agent_engine/api/app.py#L337).

## Related (noticed while investigating, not blocking this fix)

`_log_call(logging.WARNING, "tool call failed", call, ms=latency_ms)` at [nodes.py:354](https://github.com/extra-org/extra/blob/main/src/agent_engine/engine/langgraph/nodes.py#L354) does not pass `error=` into the log fields, so that particular log line never shows the exception text — only the (separate) `on_tool_error` hook path does. Worth a one-line fix in the same area, but it's a different log line than the client-facing leak above.

rishu685:
Closes #69

### Summary of Changes
- Sanitizes `used_tools[].error` at the API boundary (`routes.py` and `app.py`) to `"Tool execution failed"` so internal Python/network exception details are not exposed to clients.
- Preserves full exception logging in server-side logs and LLM model prompts.
- Adds unit tests in `tests/agent_manager/test_api.py`.

### DIFF
--- a/src/agent_engine/api/app.py
+++ b/src/agent_engine/api/app.py
@@ -118,6 +118,14 @@ class ToolRecord(BaseModel):
     error: str | None = None
 
 
+def _client_tool_record(t: Any) -> ToolRecord:
+    fields = dataclasses.asdict(t)
+    if fields.get("error"):
+        fields["error"] = "Tool execution failed"
+    return ToolRecord(**fields)
+
+
+
 class PendingApprovalModel(BaseModel):
     """Sanitized pending-approval payload returned to the client/UI."""
 
@@ -255,7 +263,8 @@ def create_app(
             system_name=result.system_name,
             answer=result.answer,
             visited=result.visited,
-            used_tools=[ToolRecord(**dataclasses.asdict(t)) for t in result.used_tools],
+            used_tools=[_client_tool_record(t) for t in result.used_tools],
+
             run_id=run_id,
             status=result.status,
             pending_approval=_pending_model(result.pending_approval),
@@ -334,7 +343,8 @@ def create_app(
             system_name=result.system_name,
             answer=result.answer,
             visited=result.visited,
-            used_tools=[ToolRecord(**dataclasses.asdict(t)) for t in result.used_tools],
+            used_tools=[_client_tool_record(t) for t in result.used_tools],
+
             run_id=run_id,
             status=result.status,
             pending_approval=_pending_model(result.pending_approval),
--- a/src/agent_engine/api/app.py
+++ b/src/agent_engine/api/app.py
@@ -125,11 +125,11 @@ def _client_tool_record(t: Any) -> ToolRecord:
     return ToolRecord(**fields)
 
 
-
 class PendingApprovalModel(BaseModel):
     """Sanitized pending-approval payload returned to the client/UI."""
 
     run_id: str
+
     approval_id: str
     agent_id: str
     tool_name: str
--- a/src/agent_engine/engine/langgraph/nodes.py
+++ b/src/agent_engine/engine/langgraph/nodes.py
@@ -351,7 +351,8 @@ class AgentNode:
         """
         error = str(exc)[:200]
         used_tools.append(self._usage(call, "failed", error=error))
-        self._log_call(logging.WARNING, "tool call failed", call, ms=latency_ms)
+        self._log_call(logging.WARNING, "tool call failed", call, ms=latency_ms, error=error)
+
         await self._hook_manager.run_on_tool_error(
             current_run_context.get(),
             self._call_context(call, "failed", latency_ms, error=error),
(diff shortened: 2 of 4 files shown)


Asaf-prog:
Thanks — the overall direction is correct, and sanitizing at the API boundary is the right approach.

The `agent_manager` normal and streaming responses are now covered, and the engine `/invoke` and approval-decision responses also use the sanitized `ToolRecord`.

There is still one client-facing path that appears to expose the raw tool error: `agent_engine/api/app.py`’s `/stream` endpoint.

It currently serializes the complete `RunStreamEvent` using:

```python
payload = {
    k: v
    for k, v in dataclasses.asdict(event).items()
    if v is not None
}
```

Since `RunStreamEvent.used_tools` contains the internal `ToolUsageRecord` objects, a final streaming event can still include the original exception text.

Please sanitize `used_tools` in this path as well, preferably through the same client-facing serialization boundary, and add a regression test for the direct engine `/stream` endpoint. It would also be useful to cover the engine `/invoke` response since that file was changed but the current tests only exercise `agent_manager`.

Once that path is covered, the fix looks good to me.

Asaf-prog:
I’m approving this PR as-is since it addresses the main issue correctly.

Please open a separate follow-up PR to sanitize the direct agent_engine /stream path as well and add the relevant regression test there.

Asaf-prog (approved):

PR #85, opened 2026-08-02
generate: report tools declared but not referenced by any agent

### LINKED ISSUE #67: generate: silently omits tools declared but not referenced by any agent

LuisMend12:
## Summary
- `Generator.generate()` only looked at `tool_ids` collected by walking the agent graph (`_collect()`), so a tool declared in the top-level `tools:` registry but never added to any agent's `tools:` list was invisible in `agentctl generate` output — no stub (correct, nothing uses it) but no mention either. A forgotten reference looked identical to an intentional one.
- Added `SystemSpec.declared_tools`, populated from the full top-level `tools:` registry during YAML parsing (previously this registry was only used transiently inside the parser to resolve per-agent tool references, then discarded).
- `Generator.generate()` now diffs `declared_tools` against the referenced `tool_ids` and reports the difference on a new `GenerateResult.ignored` field.
- `agentctl generate` surfaces it:
  ```
    ignore  add_new_user  (declared but not referenced by any agent)
  ```

Fixes #67

## Test plan
- [x] New test `tests/parsers/test_declared_tools.py` — confirms the parser populates `SystemSpec.declared_tools` (ids + descriptions) from real YAML, using the exact repro from the issue
- [x] New tests in `tests/generate/test_plugins_manifest.py` — `Generator.generate()` reports an unreferenced declared tool in `result.ignored` (and not in `created`/no stub file), and reports nothing when all declared tools are referenced
- [x] Ran `tests/parsers`, `tests/generate`, `tests/core` — all pass except 5 pre-existing failures unrelated to this change (a Windows-only path-separator assertion bug: `result.created.count("plugins/plugins.toml")` compares a forward-slash literal against `str(Path.relative_to(...))`, which yields backslashes on Windows — reproduces identically on `main` before this change)

### DIFF
--- a/src/agent_engine/core/spec.py
+++ b/src/agent_engine/core/spec.py
@@ -168,3 +168,6 @@ class SystemSpec:
     hooks: HooksConfig = field(default_factory=HooksConfig)
     plugins: PluginsConfig = field(default_factory=PluginsConfig)
     execution: ExecutionPolicy = field(default_factory=ExecutionPolicy)
+    # The full top-level `tools:` registry, independent of which agents
+    # reference them. Lets `generate` report tools declared but unused.
+    declared_tools: tuple[ToolSpec, ...] = field(default_factory=tuple)
--- a/src/agent_engine/generate/generator.py
+++ b/src/agent_engine/generate/generator.py
@@ -15,6 +15,7 @@ from agent_engine.generate.manifest import (
 class GenerateResult:
     created: list[str] = field(default_factory=list)
     skipped: list[str] = field(default_factory=list)
+    ignored: list[str] = field(default_factory=list)
 
 
 class Generator:
@@ -44,6 +45,10 @@ class Generator:
             self._write(
                 tools_dir / f"{tool_id}.py", _tool_stub(tool_id, description), result, base_dir
             )
+
+        result.ignored.extend(
+            t.id for t in spec.declared_tools if t.id not in tool_ids
+        )
 
         resolvers_dir = base_dir / "plugins" / "resolvers"
         resolvers_dir.mkdir(parents=True, exist_ok=True)
--- a/src/agent_engine/parsers/yaml/parser.py
+++ b/src/agent_engine/parsers/yaml/parser.py
@@ -658,6 +658,10 @@ class YAMLParser(Parser):
             hooks=_build_hooks(data.get("hooks")),
             plugins=_build_plugins(data.get("plugins")),
             execution=_build_execution(data.get("execution")),
+            declared_tools=tuple(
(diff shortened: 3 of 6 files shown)


Asaf-prog:
Thanks for the contribution — the implementation direction is correct, and preserving the full top-level tool registry as `declared_tools` is a clear approach.

However, PR #74 is already open for the same issue and implements the same behavior, including the `ignored` result, CLI output, and regression coverage.

PR #74 also handles the ignored-only CLI case more accurately. In this PR, the command can currently print:

```text
ignore add_new_user (declared but not referenced by any agent)
✓ Nothing to generate — all stubs already exist.
```

That summary is misleading because the ignored tool's stub does not exist by design. PR #74 uses a more accurate summary and includes an end-to-end CLI regression test.

To avoid maintaining and merging two competing implementations for #67, I think we should continue with #74 and close this PR as a duplicate.

The `declared_tools` naming used here is clearer, though, and is worth carrying over to #74.

Thank you again for taking the time to work on this.

LuisMend12:
Totally agree, thanks for checking — didn't see #74 was already in flight. The ignored-only summary bug you caught there is a good catch, glad it's covered. Appreciate you calling out the `declared_tools` naming, hope it's useful if you revisit that field later.

PR #90, opened 2026-08-08
Chore/refactor engine

Asaf-prog:


src/agent_manager/application/service.py
```diff
@@ -210,53 +210,62 @@ async def complete_turn(
...
+                conversation_id=turn.session_id,
+                user_id=turn.user_id,
+            ),
+        ):
+            if event.type == "final":
+                final = event
+
+            yield event
+
+        if final is None:
```
AmitAvital1: **Blocking — the assistant turn is lost when the stream fails after `final`.**

Without the old `except Exception: if final is None: raise`, any engine failure after the `final` event skips `append_message`, so the answer disappears from history and from the next turn's context.

This is reachable for real, not just via the test double: an `on_run_end` hook with `failure_policy != "warn"` raises inside `RunLifecycle.succeed` *after* `on_completed()` already emitted `final`, and `_stream_graph` turns it into `channel.abort(exc)`.

Propagating the failure is right; losing the message isn't. Persist first, then let it propagate:

```python
try:
    async for event in self._engine.stream(...):
        if event.type == "final":
            final = event
        yield event
finally:
    if final is not None:
        await self._persist_assistant_turn(turn, final)
```

`finally` also covers client disconnect, which is what you want. Awaiting there after `GeneratorExit` is fine as long as you don't yield, and this doesn't.

src/agent_manager/api/routes.py
```diff
@@ -167,7 +161,6 @@ async def stream_message(
     except StopAsyncIteration:
         first = None
 
-
     async def event_source() -> AsyncIterator[str]:
```
AmitAvital1, line 168: **Blocking (line 173, `error: str(exc)`).** That line is unchanged, but the service change above makes it reachable for engine-internal failures, so raw exception text now goes to the client. The new test asserts exactly that (`assert "cleanup after final" in text`) while `test_tool_error_text_is_sanitized_in_stream_message` asserts the opposite policy for the same channel. Log `exc` server-side and send a generic message, the way `_BUDGET_EXCEEDED_DETAIL` does.

src/agent_engine/runs/repository.py
```diff
@@ -0,0 +1,43 @@
+"""Abstract persistence contract for run lifecycle state."""
+
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+
+from agent_engine.approvals.models import RunRecord, RunStatus
+
+
+class RunRepository(ABC):
```
AmitAvital1: `RunRepository` is an `ABC` here, but `ApprovalRepository` and `ToolExecutionRepository` one file over are still `@runtime_checkable Protocol` — two port styles in the same layer. It also narrows the contract: an existing duck-typed run repository now has to subclass. `approvals/repository.py` still says "structural contracts" too. Suggest keeping `Protocol` for consistency.

pyproject.toml
```diff
@@ -125,3 +125,8 @@ ignore_missing_imports = true
 testpaths = ["tests"]
 addopts = "-q"
 asyncio_mode = "auto"
+
+[dependency-groups]
```
AmitAvital1: `[project.optional-dependencies].dev` already declares `ruff>=0.6`. Two dev groups with two different ruff constraints means `make format-check` passes or fails depending on which one was installed. Bump the existing extra and drop this group.

src/agent_engine/approvals/models.py
```diff
@@ -1,17 +1,3 @@
...
-* ``thread_id``    — LangGraph persistence / checkpoint identifier.
-* ``approval_id``  — one pending approval request.
-* ``tool_call_id`` — the agent's requested tool call.
-* ``execution_id`` — a single actual execution attempt (idempotency key).
-
-State transitions are explicit and validated; illegal transitions raise so a
-completed or rejected run can never silently re-run.
-"""
-
 from __future__ import annotations
```
AmitAvital1, line 15: This docstring was the only place the five identifiers (`run_id` / `thread_id` / `approval_id` / `tool_call_id` / `execution_id`) and their separation were written down, and the `_RUN_TRANSITIONS` comment went with it. Nothing in the PR replaces either, and neither is related to the refactor — please restore.

src/agent_engine/approvals/repository.py
```diff
@@ -1,37 +1,23 @@
-"""Persistence contracts for runs, approvals, and execution idempotency.
-
-The engine depends only on these ``Protocol`` interfaces (Dependency Inversion);
-the in-memory implementations here are the default and back local development
-and tests. A production deployment supplies a shared, database-backed
-implementation with the *same* contract so multi-pod resume works — the
-distributed-safe atomic claim maps to a conditional ``UPDATE ... WHERE
-status = 'pending'`` (compare-and-set).
+"""Persistence contracts for approvals and execution idempotency.
```
AmitAvital1, line 1: The dropped sentence carried the implementation hint for the future SQL adapter: the atomic claim maps to `UPDATE ... WHERE status = 'pending'`. `docs/HUMAN_IN_THE_LOOP.md` gained the equivalent note for runs but not for approvals, so that guidance is now gone. Worth keeping.

src/agent_engine/runs/in_memory.py
```diff
@@ -0,0 +1,51 @@
...
+            return True
+
+    async def get(self, run_id: str) -> RunRecord | None:
+        async with self._lock:
+            return self._runs.get(run_id)
+
+    async def transition_if_allowed(self, run_id: str, target: RunStatus) -> bool:
+        async with self._lock:
+            record = self._runs.get(run_id)
+            if record is None or not can_run_transition(record.status, target):
```
AmitAvital1: `can_run_transition` and then `record.transition(target)`, which calls `ensure_run_transition` on the same table again. Either wrap `record.transition()` in `except InvalidStateTransition: return False`, or keep the guard and drop the second check.

src/agent_engine/engine/langgraph/stream_channel.py
```diff
@@ -0,0 +1,45 @@
...
+
+    async def events(self) -> AsyncIterator[RunStreamEvent]:
+        """Yield events until the producer closes the channel, re-raising a
+        producer failure in the consumer's task."""
+        while True:
+            item = await self._queue.get()
+            if item is None:
+                return
+            if isinstance(item, BaseException):
+                raise RuntimeError(str(item)) from item
```
AmitAvital1: `raise RuntimeError(str(item)) from item` erases the exception type, so callers can't map it. Pre-existing behaviour, but moving it into a new 45-line file is the cheap moment to `raise item` instead — and it makes the sanitization fix in `routes.py` easier to do properly.

Makefile
```diff
@@ -39,9 +39,14 @@ sync-ai: generate-ai ## Alias for generate-ai (older name, kept for docs compati
...
 sync-skills: generate-ai ## Alias for generate-ai (older name, kept for docs compatibility).
 
-format: ## Auto-format the codebase (ruff format).
+format-check: ## Check code formatting without modifying files.
+	ruff format --check $(SRC) $(TESTS)
+
+format: ## Auto-format the codebase.
 	ruff format $(SRC) $(TESTS)
 
+check: format-check lint typecheck test generate-check ## Quality gate.
```
AmitAvital1: `check` now sits above the targets it depends on, and there's a leftover double blank line further down (57-58). Also `CLAUDE.md` still describes `make check` as "lint + typecheck + test" — it runs `format-check` first now.

tests/core/test_plugin_stub_scan.py
```diff
@@ -71,9 +71,7 @@ def write(base: Path, rel: str, text: str) -> None:
 
 
 STUB_TOOL = (
```
AmitAvital1: Blanket `ruff format` collapsed these multi-line source fixtures into one string with embedded `\n`, which reads worse than the implicit-concatenation form. Not worth reverting on its own, but it's an argument for landing "apply ruff formatting across the codebase" as its own PR — right now it touches ~15 files unrelated to the engine refactor and makes this diff harder to read than it needs to be.

AmitAvital1 (commented):
Reviewed the whole branch. The refactor is sound and I verified it locally: `ruff format --check`, `ruff check` and `mypy src` are clean, and the only 3 failing tests (`tests/cli/test_run_persistence.py`, missing `conversation_messages` table) fail identically on `main` — environment, not this PR.

Good fixes in here worth calling out: `suppress(ValueError)` around the contextvar resets fixes a real crash when an async generator is finalized in another context; `create_if_absent` stops `create()` from overwriting a properly-named run record with `system_name=""`; `transition_if_allowed` removes the read-then-write race; and cancelling abandoned streams closes a genuine leak.

Two blocking findings on the manager streaming path, plus a few smaller ones — all inline.

src/agent_engine/approvals/models.py
```diff
@@ -60,8 +73,20 @@ class ApprovalStatus(StrEnum):
 }
 
 
+def can_run_transition(current: RunStatus, target: RunStatus) -> bool:
```
AmitAvital1: Nit — this is dead public API now. Removing the double validation in `in_memory.py` took away its only external caller, so it's referenced solely by `ensure_run_transition` four lines below, in this same module, and nothing in `tests/` touches it either.

Inline it and drop the function plus its docstring:

```python
def ensure_run_transition(current: RunStatus, target: RunStatus) -> None:
    if target not in _RUN_TRANSITIONS[current]:
        raise InvalidStateTransition("run", current.value, target.value)
```

If a shared adapter later wants a non-raising check, it's two lines to bring back.

AmitAvital1 (commented):
Re-reviewed after the fixes — all ten points are addressed, and several went further than I asked. Verified on `b8fd33f`: `ruff format --check`, `ruff check` and `mypy src` clean, same 3 pre-existing `test_run_persistence.py` failures as `main`.

Worth calling out:

- Extracting `_persist_assistant_turn` is better than what I suggested — `complete_turn` now shares it, so the duplicated `ConversationMessage` construction is gone too.
- `send_message` was leaking `str(exc)` in its 500 as well; sanitizing that too (with `test_send_sanitizes_engine_failure`) was beyond the ask and correct.
- `raise item` is verified end-to-end with a distinct exception type rather than a message string. Good test.
- `dedent()` for the plugin fixtures is better than the pre-PR implicit concatenation — readable *and* formatter-stable.
- Docs churn fully reverted: net diff vs `main` for `docs/` is now only the files with real content changes.

On the `finally` in `service.stream`: I was unsure whether an await there survives a *cancelled* task (Starlette's disconnect path, not `aclose()`) — that's the case the engine reaches for `asyncio.shield` for. Built a repro with a repository that actually suspends on write and it persists correctly: cancellation is delivered once and the `finally` finishes its awaits. No shield needed, no issue.

One leftover nit inline. Not a merge blocker.

AmitAvital1 (approved):
Approved. `can_run_transition` inlined and gone — no stale references anywhere in the tree.

Verified on `c234f56`:

- `pytest` — 610 passed, 3 failed. The 3 are `tests/cli/test_run_persistence.py` (`no such table: conversation_messages`) and fail identically on `main`. Environment/alembic, not this branch.
- `ruff format --check` — 186 files already formatted
- `ruff check` — all checks passed
- `mypy src` — no issues in 106 source files

All 11 review points addressed across three rounds. Recap of what actually changed in the code as a result:

**Correctness** — the assistant turn is no longer lost when the engine fails after `final` (`service.py`, now via a shared `_persist_assistant_turn` that `complete_turn` also uses, so the duplicated `ConversationMessage` construction went away). Producer exception types survive the stream transport (`raise item`). Run transitions validate once, under the adapter lock.

**Security** — internal exception text no longer reaches clients on either the SSE or the JSON path; failures are logged server-side and returned as a stable generic error. `send_message` was leaking too and got fixed alongside the reported one.

**Consistency** — `RunRepository` back to `Protocol`, matching its sibling ports. Single ruff constraint. Make targets ordered after their dependencies, `CLAUDE.md` in sync with the real gate.

**Docs** — the five-identifier spec and the SQL compare-and-set guidance restored; unrelated formatter churn in `docs/` fully reverted, so the net diff there is only the content that actually changed.

The engine refactor underneath was solid from the start: `RunLifecycle` / `StreamChannel` / `runs.repository` are a real improvement over the 300-line `stream()`, and the contextvar-reset fix, `create_if_absent`, and abandoned-stream cancellation each close a genuine bug. Test coverage for the new lifecycle is behavior-focused and pins the ordering guarantees the code depends on.

Nice work on the turnaround.

PR #94, opened 2026-08-11
feat(manager)!: verified identity for the conversation API

AmitAvital1:
## Summary

- **Replaces `X-Agent-Chat-User`** — a header the browser set for itself, so anyone could send another user's id and read their conversations — with a signed token verified against a host- or manager-issued key.
- **Two token sources, one verifier**: the host's own session token (`host_token` mode — zero host code when the manager is proxied under the host's origin), or a token the host mints for us (`mint` mode). Products with no login get a manager-issued visitor pass automatically.
- **Anonymous → account merge**: a visitor's conversations move onto their real account the moment they sign in, verified and idempotent (`POST /auth/link`), via a conditional `UPDATE` so two tabs signing in at once can't double-claim the same conversations.
- **Extensible verification**: `TokenVerifier` takes its algorithms from a `KeySource` protocol rather than naming them, so RS256/ES256/JWKS is a new implementation later, not a rewrite.
- Engine boundary untouched — `agent_engine` has zero knowledge of tokens; the manager still hands it `message + history + RunContext(user_id=...)` per request.

## Why

`conversation_users`/`conversation_sessions` already enforced per-user ownership — the only gap was that the id being compared was never proven. `curl -H 'X-Agent-Chat-User: alice' .../conversations` returned Alice's conversations to anyone.

## Breaking change

`X-Agent-Chat-User` and the widget's `user=` attribute are removed. Every conversation route now requires a verified token (`401` without one). Conversations already stored under the old unverified ids become unreachable, not exposed — the safe direction, and expected on data this is pre-release.

## Test plan

- [x] `make check` (lint, mypy, full pytest) — green on CI
- [x] `npm run build:widget`, `typecheck:widget`, `test:widget` (unit, incl. token de-dup and visitor-pass-handoff regression tests)
- [x] `npm run test:widget:e2e` — 24/24 Playwright
- [x] Repository contract test (`link_anonymous_user`) run identically against the in-memory and SQL backends
- [ ] **Not** verified end-to-end against a real Open WebUI instance — see the correction below
- [ ] PostgreSQL — untested in this repo (pre-existing gap, not introduced here)

---

## Correction to this description

Two claims in the original test plan were wrong. Recording them here rather than
quietly editing them out, since this PR is already merged and the history should
be accurate.

**1. "Manually verified end-to-end against a real Open WebUI instance."** This was
not true. The chat worked during that testing, and I took that as proof the
identity path worked. It did not: every request was falling back to an anonymous
visitor pass, for two independent reasons found afterwards —

- the `sub` bug caught in review (@Asaf-prog), which rejected Open WebUI's real
  token shape outright; and
- a `token-url` that resolved to the frontend dev server rather than the backend,
  returning 404.

Both failed *silently*, because the widget degrades to an anonymous pass when it
cannot obtain a host identity. The database told the real story: `anon:` rows
where `ext:` rows should have been. #101 makes that failure mode loud so this
cannot happen the same way again.

**2. "3 pre-existing `test_run_persistence.py` failures verified present on
`origin/main`."** They were present locally, but for an unrelated Python 3.13
environment problem that masked the truth. On CI (Python 3.11) two of those tests
failed *because of this change* — they asserted the raw `local-user` id, which is
now digested. Fixed in 0f6bb5b.

What *was* verified after the fact, empirically: the route responds, the secrets
match, a token in Open WebUI's exact shape is accepted, and it resolves to
`ext:<digest>` with the real user id — not an anonymous pass. The full
browser-with-a-logged-in-session path remains unverified.


Asaf-prog:
Blocking this for now because the configurable claim mapping is currently inconsistent with token verification.

`ClaimMapping` correctly allows different hosts to define where the user id lives, for example `sub`, `id`, or another claim via `AGENT_AUTH_CLAIM_USER_ID`.

However, `TokenVerifier` still always requires `sub` to exist before `ClaimMapping` runs.

That means a valid host configuration such as:

```bash
AGENT_AUTH_CLAIM_USER_ID=id
```

with a token containing:

```json
{
  "id": "user-123"
}
```

will still be rejected because `sub` is missing.

The separation should be:

* `TokenVerifier` validates that the token is authentic and valid: signature, expiry, issuer, audience, etc.
* `ClaimMapping` validates and extracts the configured user-id claim.

Because the current implementation breaks one of the main advertised authentication configurations, I don't think this should be approved until the unconditional `sub` requirement is removed or made specific to the identity source that actually requires it.

AmitAvital1:
Confirmed — genuine bug, thanks for catching it. Repro'd exactly: a token shaped like Open WebUI's real ones (`{"id": "u_8412", "jti": ..., "iat": ..., "exp": ...}` — no `sub` at all) was rejected before `ClaimMapping` ever got to look at it.

Root cause: `TokenVerifier._required_claims()` hardcoded `sub` into PyJWT's `require` option unconditionally, disconnected from `AGENT_AUTH_CLAIM_USER_ID`/`ClaimMapping`. Fixed in 4697e1a — `TokenVerifier` no longer knows or cares about `sub`; it only enforces what verification itself needs (signature, expiry, issuer/audience). `ClaimMapping` was already checking and raising correctly on the *configured* claim — that's now the only place identity-claim presence is enforced.

Also fixed why nothing caught this: the test helper defaulted every token to include `sub="alice"` regardless of what claim was actually under test. Added a regression test using a raw payload shaped exactly like Open WebUI's real tokens; verified it fails on the pre-fix code with the exact error you reported, passes after.

Asaf-prog (approved):

PR #96, opened 2026-08-13
fix(config): rename AGENT_env vars to EXTRA_ with backward-compat deprecation warning

### LINKED ISSUE #95: Rename AGENT_* environment variables to EXTRA_* to match the product brand
## Context

`CLAUDE.md` is explicit that the product/brand name is **Extra**; `agent-engine`/`agent-manager` are only the PyPI package and console-script names, never meant to be user-facing.

Every runtime setting, though, is prefixed `AGENT_` — `AGENT_DB_BACKEND`, `AGENT_DB_URL`, and now (from #94) the whole `AGENT_AUTH_*` family (`AGENT_AUTH_MODE`, `AGENT_AUTH_SECRET`, `AGENT_AUTH_COOKIE`, `AGENT_AUTH_CLAIM_*`, etc.).

## Why this is worth fixing

- **Brand mismatch.** An admin configuring "Extra" will reasonably look for `EXTRA_*`, not `AGENT_*` — the env-var surface leaks an internal package name into the one place customers actually type things.
- **Collision risk in enterprise environments.** "agent" is a heavily overloaded word in infra (monitoring agents, security agents, etc.). `AGENT_AUTH_MODE` sitting next to e.g. `DATADOG_AGENT_*` in a shared secrets manager or ConfigMap is genuinely easy to misread or collide with.

## Why it's not part of #94

#94 only *adds* new settings; it followed the `AGENT_DB_*` naming that already shipped rather than introducing a second, inconsistent prefix (`EXTRA_AUTH_*` next to `AGENT_DB_*`) inside one PR. A rename is a deliberate, product-wide decision, not something to fold into a feature branch.

## Suggested approach

- Introduce `EXTRA_*` as the primary name for every setting.
- Keep the old `AGENT_*` names working as deprecated aliases for at least one release (warn on use, don't break existing deployments outright), then drop them.
- Update `.env.example`, `examples/*/.env.example`, and the docs (`docs/identity.mdx`, `docs/quickstart.mdx`, etc.) in the same pass.

rishu685:
Closes #95.

### Summary
Renames AGENT_DB_BACKEND and AGENT_DB_URL environment variables to EXTRA_DB_BACKEND and EXTRA_DB_URL in src/agent_manager/config.py to match the Extra product brand name.

### Fixes & Changes
1. src/agent_manager/config.py:
   - Renamed agent_db_backend and agent_db_url settings fields to extra_db_backend and extra_db_url.
   - Added`@model_validator(mode="before") method _backfill_deprecated_agent_vars using module-level constants _DEPRECATED_ENV_VARS.
   - Emits a DeprecationWarning when legacy AGENT_DB_ environment variables are present and falls back to them when EXTRA_DB_ is absent.
   - Preserves precedence for EXTRA_DB_ if both old and new environment variables are set.
   - 
2. tests/agent_manager/test_config.py:
   - Added unit tests for EXTRA_DB_ environment variable parsing.
   - Added tests verifying AGENT_DB_ deprecation warnings and fallback behavior.
   - Added test verifying EXTRA_DB_ precedence over deprecated AGENT_DB_.
3. tests/cli/test_run_persistence.py:
   - Updated CLI persistence tests to use EXTRA_DB_BACKEND and EXTRA_DB_URL.

### Verification
- pytest: 572 passed
- ruff check: All checks passed
- mypy: Success (no issues found)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -17,11 +17,11 @@ ANTHROPIC_API_KEY=your-anthropic-key-here
 #              Needs the manager served from the host's origin.
 #   mint       the host mints a short-lived token for us from a shared secret.
 #
-# AGENT_AUTH_MODE=host_token
-# AGENT_AUTH_SECRET=copy-the-hosts-jwt-secret-here-32-chars-min
-# AGENT_AUTH_CLAIM_USER_ID=id
-# AGENT_AUTH_COOKIE=token
+# EXTRA_AUTH_MODE=host_token
+# EXTRA_AUTH_SECRET=copy-the-hosts-jwt-secret-here-32-chars-min
+# EXTRA_AUTH_CLAIM_USER_ID=id
+# EXTRA_AUTH_COOKIE=token
 #
-# Signs the visitor passes we issue. Derived from AGENT_AUTH_SECRET when unset;
+# Signs the visitor passes we issue. Derived from EXTRA_AUTH_SECRET when unset;
 # set it explicitly in production so passes survive a secret rotation.
-# AGENT_AUTH_ANONYMOUS_SECRET=a-separate-secret-32-characters-minimum
+# EXTRA_AUTH_ANONYMOUS_SECRET=a-separate-secret-32-characters-minimum
--- a/docs/api.mdx
+++ b/docs/api.mdx
@@ -97,7 +97,7 @@ Identity arrives as a signed JWT, in either of two places:
 
 - `Authorization: Bearer <token>` — a token the host app minted, or a visitor
   pass the manager issued.
-- The host's own session cookie, named by `AGENT_AUTH_COOKIE`, when the manager
+- The host's own session cookie, named by `EXTRA_AUTH_COOKIE`, when the manager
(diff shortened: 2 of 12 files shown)


Asaf-prog:
Thanks for working on this. The direction looks good.

Before merging, I think this PR should fully cover the scope of issue #95. Since #94 has now added the `AGENT_AUTH_*` configuration family, please rebase/update this PR against the latest `main` and rename those variables to `EXTRA_AUTH_*` as well.

Please also update the relevant `.env.example` and documentation references so the public configuration surface is consistently using `EXTRA_*`.

Once that is updated, this should be good to go.

AmitAvital1:
Hi we just merge https://github.com/extra-org/extra/pull/94 which include more env vars (also don't forget change official docs and examples if they use)

rishu685:
Thanks for the review feedback @Asaf-prog and @AmitAvital1! I've updated the PR accordingly and pushed the latest changes.

Asaf-prog (approved):

PR #98, opened 2026-08-14
Fix/hitm approval actions

Asaf-prog:
## Summary

Adds complete end-to-end Human-in-the-Loop approval support to the agent manager and web widget.

Previously, the engine could pause for approval, but the manager and widget could not reliably submit a decision, resume the same run, persist its result, or support session-level approval.

### Engine

- Added a resumable approval capability through the `ApprovalEngine` protocol.
- Added safe approval authorization using the caller’s user and session identity.
- Added recovery for already-processed approval decisions.
- Preserved cumulative token usage across approval/resume execution legs.
- Added support for session-scoped approval grants.

### Agent Manager

- Added an approval decision endpoint:

  `POST /conversations/{conversation_id}/runs/{run_id}/approvals/{approval_id}/decision`

- Exposed pending approval metadata to API and SSE clients.
- Resumed suspended runs after approval decisions.
- Persisted final assistant responses idempotently after resumed execution.
- Handled repeated decision requests safely.
- Sanitized internal approval and engine errors before returning them to clients.

### Web Widget

Added three approval actions:

| Action | Behavior |
| --- | --- |
| Approve | Allows this specific tool execution |
| Deny | Rejects this specific tool execution |
| Approve for this session | Allows the execution and remembers the relevant approval scope for the current session |

Additional UI behavior:

- Displays an approval card when execution is paused.
- Prevents duplicate approval requests.
- Disables the composer while approval is pending.
- Restores controls when a decision fails.
- Supports chained approval requests.
- Prevents repeated prompts after session-level approval.
- Keeps internal backend errors hidden from users.

### Documentation

Updated the HITL and widget documentation to describe:

- approval lifecycle and resume behavior;
- approval decision semantics;
- session-level approval scope;
- manager API integration;
- widget behavior and failure handling.

## Validation

- `make check` — passed
  - Ruff formatting and lint passed
  - Mypy passed
  - 681 Python tests passed
- `npm run typecheck:widget` — passed
- `npm run test:widget` — passed
- `npm run test:widget:e2e` — 29 tests passed

image

### DIFF
--- a/docs/widget.mdx
+++ b/docs/widget.mdx
@@ -152,16 +152,17 @@ The widget connects to the streaming endpoint (`POST /conversations/{id}/message
 ## Tool approvals
 
 When a tool requires human approval, the widget pauses the current run and
-shows three actions:
+shows the tool's human-readable description and three actions:
 
 - **Approve** — allow only this tool invocation.
 - **Deny** — reject this invocation without running the tool.
 - **Approve for this session** — allow this invocation and stop asking for the
   same scoped tool during the current conversation.
 
-The decision resumes the existing run; it does not submit the user message a
-second time. Session approval remains scoped by the backend to the current
-system, identity, conversation, agent, and provider-qualified tool.
+The decision resumes the existing run through the conversation approval API; it
+does not resend the user's message. Hosts must keep the same authenticated
+conversation identity for the decision. Afterward, the run either completes or
+shows the next approval request.
 
 ---
 
--- a/docs/widget.mdx
+++ b/docs/widget.mdx
@@ -149,6 +149,20 @@ Only safe metadata is exposed — no internal reasoning, no prompt content.
 
 The widget connects to the streaming endpoint (`POST /conversations/{id}/messages/stream`) and renders tokens as they arrive. If streaming fails before a usable answer, it falls back to the standard endpoint automatically.
 
+## Tool approvals
+
+When a tool requires human approval, the widget pauses the current run and
+shows three actions:
+
+- **Approve** — allow only this tool invocation.
+- **Deny** — reject this invocation without running the tool.
+- **Approve for this session** — allow this invocation and stop asking for the
+  same scoped tool during the current conversation.
+
+The decision resumes the existing run; it does not submit the user message a
+second time. Session approval remains scoped by the backend to the current
+system, identity, conversation, agent, and provider-qualified tool.
+
 ---
 
 ## Session recovery
--- a/src/agent_engine/approvals/manager.py
+++ b/src/agent_engine/approvals/manager.py
@@ -215,12 +215,30 @@ class ApprovalManager:
     async def get_approval(self, run_id: str, approval_id: str) -> ApprovalRecord:
         return await self._load_for_run(run_id, approval_id)
 
+    async def get_authorized(
+        self,
+        *,
+        run_id: str,
+        approval_id: str,
+        caller_user_id: str | None = None,
+        caller_auth_ref: str | None = None,
+    ) -> ApprovalRecord:
+        """Return an approval only when the caller matches its stored scope."""
+        record = await self._load_for_run(run_id, approval_id)
+        self._ensure_authorized(
+            record,
+            caller_user_id=caller_user_id,
+            caller_auth_ref=caller_auth_ref,
+        )
+        return record
+
     async def claim(
         self,
         *,
         run_id: str,
         approval_id: str,
         caller_user_id: str | None = None,
+        caller_auth_ref: str | None = None,
     ) -> ApprovalRecord:
         """Validate and atomically claim an approval for resume.
 
@@ -229,9 +247,13 @@ class ApprovalManager:
         :class:`ApprovalAlreadyProcessed`. On success the run and approval both
         move to RESUMING.
         """
-        record = await self._load_for_run(run_id, approval_id)
-        if record.authorized_user_id is not None and record.authorized_user_id != caller_user_id:
-            raise UnauthorizedApprover(approval_id)
+        record = await self.get_authorized(
+            run_id=run_id,
+            approval_id=approval_id,
+            caller_user_id=caller_user_id,
+            caller_auth_ref=caller_auth_ref,
+        )
+
         try:
             claimed = await self._approvals.claim(approval_id)
         except InvalidStateTransition as exc:
@@ -255,6 +277,18 @@ class ApprovalManager:
             tool_call_id=claimed.tool_call_id,
         )
         return claimed
+
+    @staticmethod
+    def _ensure_authorized(
+        record: ApprovalRecord,
+        *,
+        caller_user_id: str | None,
+        caller_auth_ref: str | None,
+    ) -> None:
+        if record.authorized_user_id is not None and record.authorized_user_id != caller_user_id:
+            raise UnauthorizedApprover(record.approval_id)
+        if record.auth_ref is not None and record.auth_ref != caller_auth_ref:
+            raise UnauthorizedApprover(record.approval_id)
 
     async def finalize(self, approval_id: str, *, approved: bool) -> ApprovalRecord:
         target = ApprovalStatus.APPROVED if approved else ApprovalStatus.REJECTED
--- a/src/agent_engine/engine/langgraph/engine.py
+++ b/src/agent_engine/engine/langgraph/engine.py
@@ -20,7 +20,7 @@ from agent_engine.approvals.coordinator import ApprovalCoordinator
 from agent_engine.approvals.decision import ApprovalDecision, parse_decision
 from agent_engine.approvals.errors import RunNotFound
 from agent_engine.approvals.manager import ApprovalManager, ToolExecutionManager
-from agent_engine.approvals.models import ApprovalRecord
+from agent_engine.approvals.models import ApprovalRecord, RunStatus
 from agent_engine.approvals.repository import (
     InMemoryApprovalRepository,
     InMemoryToolExecutionRepository,
@@ -205,6 +205,9 @@ def _pending_approval_event(system_name: str, result: RunResult) -> RunStreamEve
         agent_id=approval.agent_id,
         tool_name=approval.tool_name,
         description=approval.description,
+        provider=approval.provider,
+        server_id=approval.server_id,
+        arguments=dict(approval.arguments),
     )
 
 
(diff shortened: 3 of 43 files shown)


Asaf-prog:
image

image

src/agent_manager/api/static/widget/react/AgentChatApp.tsx
```diff
@@ -202,28 +222,90 @@ export function AgentChatApp({
...
+        }
+      } catch (error) {
+        const is4xx = error instanceof AgentChatHttpError && error.status >= 400 && error.status < 500;
+        putEntries(cid, (prev) =>
+          prev.map((entry) =>
+            entry.id === entryId
+              ? {
+                  ...entry,
+                  approvalSubmitting: false,
+                  approvalError: is4xx ? error.message : GENERIC_ERROR,
```
AmitAvital1: The composer can dead-lock here: `awaitingApproval` stays true as long as the entry has an `approval`, and the card only clears on a *successful* decision. If the run is already gone (409 already-processed, or 404 after a restart) the retry can never succeed and the conversation is stuck with no way out.

Worth clearing `approval` on a terminal 4xx and leaving just the error text.

src/agent_engine/approvals/manager.py
```diff
@@ -256,6 +278,18 @@ async def claim(
         )
         return claimed
 
+    @staticmethod
+    def _ensure_authorized(
```
AmitAvital1, line 282: This now fails closed when `auth_ref` is set and the caller passes none — so every existing `resume()` / `POST /runs/../approve` caller that doesn't send a session id starts getting 403 (which is why the three examples and `chat.py` had to change).

Right call security-wise, but it's a public-contract break; can you call it out in the PR description / changelog?

src/agent_manager/api/routes.py
```diff
@@ -60,6 +72,15 @@
     ConversationLinkRefused: (403, "a visitor cannot adopt another visitor"),
 }
 
+_APPROVAL_HTTP_ERRORS: dict[type[ApprovalError], tuple[int, str]] = {
```
AmitAvital1: This duplicates the map in `agent_engine/api/app.py:98` — two copies of the same policy will drift. One shared `approval_http_status(exc)` next to the error classes would cover both.

They already disagree, actually: the engine endpoint returns `str(exc)` as the detail while this one returns fixed strings, so the engine side is the one still leaking internals.

src/agent_manager/infrastructure/persistence/memory_repository.py
```diff
@@ -150,6 +150,22 @@ async def append_message(
...
+    async def append_message_if_absent(
+        self,
+        message: ConversationMessage,
+        *,
+        snapshot_ttl_seconds: int | None = None,
+    ) -> bool:
+        """Append once under the adapter's single-event-loop execution model."""
+        if any(
+            existing.message_id == message.message_id
+            for messages in self._messages.values()
```
AmitAvital1: This scans every message of every session to look up one id, on every assistant write. The session is right there on the message — `self._messages.get(message.session_id, [])` is enough.

AmitAvital1 (commented):
Read through the whole thing — the layering holds up, errors are sanitized consistently on the new paths, and the `append_message_if_absent` + `uuid5` idempotency is a clean way to make the retry safe. `make lint` / `make typecheck` pass here, and the affected test dirs pass too (`tests/cli/test_run_persistence.py` fails, but it fails on `main` as well — unrelated).

Four things below. Only the first one I'd consider blocking.

docs/widget.mdx
```diff
@@ -149,6 +149,20 @@ Only safe metadata is exposed — no internal reasoning, no prompt content.
 
 The widget connects to the streaming endpoint (`POST /conversations/{id}/messages/stream`) and renders tokens as they arrive. If streaming fails before a usable answer, it falls back to the standard endpoint automatically.
 
+## Tool approvals
```
AmitAvital1, line 152: To vebose.. docs should be lean accurate and focused.

src/agent_manager/api/static/widget/react/AgentChatApp.tsx
```diff
@@ -393,16 +493,78 @@ function ChatMessage({ entry }: { entry: MessageEntry }) {
...
+  approval: PendingApproval;
+  submitting: boolean;
+  error?: string;
+  onDecision: (decision: ApprovalDecision) => void;
+}) {
+  return (
+    <section className="approval-card" aria-busy={submitting} aria-label="Tool approval request">
+      <p className="approval-title">Approval required</p>
+      <p className="approval-tool">{approval.tool_name}</p>
+      <p className="approval-description">{approval.description}</p>
```
AmitAvital1, line 551: This copy is developer-facing, and the widget's reader is a visitor on someone's site who has never seen the YAML. They get the raw tool id plus `Agent 'writer' wants to call tool 'send_email_v2' on server 'mail-prod'` — the agent id and server id mean nothing to them, and a tool id like `send_email_v2` or an MCP-prefixed name is worse.

We already hold this line elsewhere: `tests/e2e/widget.spec.ts` asserts the tool activity title shows no internal provider. `ToolSpec.description` is the human text that exists for exactly this — worth carrying it into the pending payload and rendering it, with the raw id as a secondary detail.

src/agent_manager/api/static/widget/react/useConversation.ts
```diff
@@ -93,6 +100,16 @@ export function useConversation(
     [client, replace],
   );
 
+  const decideApproval = useCallback(
```
AmitAvital1: This wrapper forwards to `client.decideApproval` verbatim, and costs an interface entry, a callback and a deps-array line for no behavior — the app can call the client directly.

Unless the intent was the opposite: `send` and `stream` earn their wrapper by recovering a stale conversation, and a decision on a vanished conversation currently just 404s (which is what leaves the card stuck — see the other comment). Either drop it or give it the same `replace` handling.

AmitAvital1 (commented):
Two more, on the user-facing side of the card.

src/agent_engine/engine/langgraph/engine.py
```diff
@@ -486,13 +518,45 @@ async def get_pending_approval(self, run_id: str) -> PendingApproval | None:
...
+            raise RunNotFound(run_id)
+        if run.status not in {RunStatus.COMPLETED, RunStatus.PENDING_APPROVAL}:
+            return None
+        ctx = RunContext(run_id=run_id)
+        snapshot = await app.aget_state(self._thread_config(ctx))
+        values = snapshot.values
+        if not isinstance(values, dict):
+            return None
+        token_usage = (run.input_tokens, run.output_tokens)
+        if run.status == RunStatus.PENDING_APPROVAL:
```
AmitAvital1, line 548: This branch never fires — recovery of a run sitting at a *second* approval always returns `None`.

`_pending_result` keys off `result["__interrupt__"]`, which only exists on the `ainvoke` return value. `aget_state(...).values` has no such key (I checked: interrupts live on `snapshot.tasks[*].interrupts`, with `snapshot.next == ('writer',)`), so it bails at the first `if not interrupts`. Effect: a client retrying a decision on a chained-approval run gets `ApprovalAlreadyProcessed` re-raised → 409 instead of the new pending approval — and in the widget that's exactly what wedges the card.

Repro against the current branch (fails on `recovered is not None`):

```python
# fake model that calls the tool twice, then answers
first = await engine.run("hi", context=RunContext(run_id="run-1"))
second = await engine.resume("run-1", first.pending_approval.approval_id, "allow once")
assert second.status == "pending_approval"  # chaining itself works
recovered = await engine.get_processed_result("run-1", first.pending_approval.approval_id)
assert recovered is not None  # -> AssertionError
```

Simplest fix is to not route the pending case through the checkpoint values at all: `get_pending_approval(run_id)` already returns the outstanding approval, so build the pending result from that plus `snapshot.values` for `visited` / `used_tools`. Worth a test for the chained case either way — the completed path has one, this one doesn't.

AmitAvital1 (commented):
Went back over the correctness side properly this time — ran the JS half too (`typecheck:widget`, `test:widget`, and all 29 Playwright e2e pass), and confirmed the committed `widget.js` bundle rebuilds byte-identical from the sources, so no drift there.

One real bug found, below.

src/agent_manager/api/static/widget/styles/styles.ts
```diff
@@ -138,6 +138,31 @@ export function styles(config: AgentChatConfig): string {
...
+      padding: 10px 11px; white-space: normal; box-shadow: 0 1px 2px rgba(0,0,0,.03); }
+    .approval-title { margin: 0; color: #3f3f46; font-size: 12.5px; font-weight: 600;
+      display: flex; align-items: center; gap: 7px; }
+    .approval-title::before { width: 6px; height: 6px; border-radius: 50%; background: #f59e0b;
+      content: ""; flex: 0 0 auto; }
+    .approval-tool { display: inline-block; margin: 6px 0 0; border: 1px solid #e4e4e7;
+      border-radius: 6px; background: #fff; color: #27272a; padding: 2px 6px;
+      font-family: ui-monospace, SFMono-Regular, Menlo, Monaco, Consolas, monospace;
+      font-size: 12px; font-weight: 500; line-height: 1.4; overflow-wrap: anywhere; }
+    .approval-description { margin: 7px 0 0; color: #71717a; font-size: 12.5px;
```
AmitAvital1, line 151: Now that this renders the tool's own description, the text is third-party for MCP tools — server-supplied, and those are routinely a few hundred chars of prose plus parameter docs written for the model.

React escapes it so there's nothing unsafe, but there's no clamp here, and a long one will push the Approve/Deny buttons out of view in a small widget. `.msg-error` a few lines down already clamps to 2 lines — same treatment (with the full text still in `title`) would keep the actions reachable.

AmitAvital1 (commented):
Went through `41fc836` and re-ran everything rather than eyeballing it. All six points are addressed, and the two that mattered are verified:

- **The recovery bug is really fixed.** My original repro (chained approvals, retry the first decision) now returns the second pending approval instead of `None` — and you added `test_retrying_first_decision_recovers_second_pending_approval` for it, which is the part I'd have asked for next. Building the pending result from `get_pending_approval` instead of the checkpoint values is the right shape; the recovered result matches the live one field for field (including `visited=[]`, same as the live path — no new inconsistency).
- **The card no longer leaks internals**, and the e2e asserts the absence of `writer` / `mail-prod` / `provider` rather than just the presence of the new text. That's the assertion that will actually catch a regression.

The rest: shared `approval_http_status` / `approval_public_message` also fixed the engine endpoint's `str(exc)` leak, which was the more interesting half of that finding; the session-id break is documented; the passthrough is gone; the memory scan is scoped to the session; and terminal-vs-retryable 4xx with `408/425/429` carved out is a more careful split than I suggested.

Verified locally: `ruff` + `mypy` clean, full pytest green, `typecheck:widget` + `test:widget` pass, all 30 Playwright e2e pass, and the committed `widget.js` still rebuilds identical to source.

One small follow-up below, non-blocking. Good from my side otherwise.

AmitAvital1 (approved):

PR #101, opened 2026-08-15
feat(widget): report identity failures instead of silently degrading

AmitAvital1:
## Summary

A configured `token-url` that could not be used was indistinguishable from a signed-out visitor: both quietly became an anonymous pass, and the chat kept working. Failures are now **classified, logged, and raised as an event**, with an opt-out of the fallback entirely.

- `agent-chat:identity-error` event — `{ reason, status, url, fellBackToAnonymous }` so a host can react itself (redirect to login, hide the widget…).
- `console.warn` for `unreachable` / `malformed`, the two that are almost always integration mistakes. Not raised for `unauthorized`, which is the ordinary signed-out case.
- `require-identity` attribute — for products where a chat with no proven user is unacceptable. Refuses to mint *or reuse* a visitor pass rather than substituting one.
- Default behaviour is unchanged, so login-less products keep working with zero configuration.

## Why

Follow-up to #94. That silence hid **two** separate real bugs during its integration, and in both cases the chat kept working so nothing looked wrong:

1. A token shape we rejected (the `sub` bug caught in review).
2. A `token-url` resolving to the dev frontend instead of the backend.

Both surfaced only by reading the database and finding `anon:` rows where `ext:` rows were expected. A warning would have surfaced either in seconds.

## Also

`TokenSource` moves to an options object — five positional parameters, three optional, had stopped being readable.

## Test plan

- [x] `ruff`, `mypy`, `pytest` — 662 passed
- [x] `npm run typecheck:widget`, `test:widget` — including new cases for each failure reason, and that `require-identity` never mints or reuses a pass
- [x] `npm run test:widget:e2e` — 24/24
- [x] Verified the new tests fail when the assertion is inverted (they genuinely assert)

### DIFF
--- a/docs/identity.mdx
+++ b/docs/identity.mdx
@@ -129,14 +129,24 @@ Someone who chats before logging in doesn't lose it. On their first
 authenticated request the widget hands their pass over and those conversations
 move onto their account. Automatic, and it only happens once.
 
-## Signing out
+## Signing in and out
+
+A normal page navigation needs nothing — the widget works out who the caller is
+on every load. A single-page app that signs a user **in or out without
+reloading** should say so, or the widget keeps the identity it already resolved:
 
 ```js
-document.querySelector("agent-chat").logout();
+const chat = document.querySelector("agent-chat");
+
+chat.refreshIdentity();  // signed in, or switched user
+chat.logout();           // signed out
 ```
 
-Clears the stored identity and the open thread. Call it when your app signs a
-user out — and, in a single-page app, when it signs one *in* without a reload.
+<Warning>
+Use `refreshIdentity()` on sign-**in**, not `logout()`. `logout()` discards the
+visitor pass, and that pass is what carries a visitor's earlier conversations
+onto their account — calling it here throws them away instead of merging them.
+</Warning>
 
 ## Settings
 
--- a/docs/identity.mdx
+++ b/docs/identity.mdx
@@ -87,6 +87,34 @@ EXTRA_AUTH_SECRET=<32+ random characters>
 
 Not logged in? Return `401` — the widget falls back to an anonymous pass.
 
+<Note>
+`token-url` is resolved like any other browser request: relative to the page.
+That is correct in production, where your app and its API share an origin. If
+your **dev** setup serves the frontend separately from your API, point it at the
+same base your own frontend uses.
+</Note>
+
+### When identity fails
+
+Anything other than a clean token is reported — a wrong URL, an endpoint
+answering with the wrong shape, an unreachable host. The widget logs a warning
+and raises an event, so a broken integration cannot quietly look like a working
+anonymous chat:
+
+```js
+document.querySelector("agent-chat").addEventListener("agent-chat:identity-error", (e) => {
+  // { reason: "unauthorized" | "unreachable" | "malformed", status, url, fellBackToAnonymous }
+  console.log(e.detail);
+});
+```
+
+By default it then continues as an anonymous visitor. If a chat with no proven
+user is unacceptable in your product, opt out of that fallback entirely:
+
+```html
+<agent-chat title="Support" token-url="/agent-chat/token" require-identity></agent-chat>
+```
+
 <Warning>
 Sign `req.user.id` from the session, never a value from the request. An endpoint
 that accepts `?user=` lets anyone get a token as anyone.
--- a/docs/identity.mdx
+++ b/docs/identity.mdx
@@ -103,7 +103,8 @@ anonymous chat:
 
 ```js
 document.querySelector("agent-chat").addEventListener("agent-chat:identity-error", (e) => {
-  // { reason: "unauthorized" | "unreachable" | "malformed", status, url, fellBackToAnonymous }
+  // { reason: "unauthorized" | "unreachable" | "malformed", status, url,
+  //   anonymousFallbackEnabled }   <- whether falling back is allowed, not that it worked
   console.log(e.detail);
 });
 ```
--- a/docs/widget.mdx
+++ b/docs/widget.mdx
@@ -61,13 +61,15 @@ Renders as a fixed panel directly in the page — no launcher button.
 | `greeting` | — | First message shown when the chat opens |
 | `mode` | `"floating"` | `"floating"` or `"inline"` |
 | `position` | `"bottom-right"` | Floating position: `"bottom-right"` or `"bottom-left"` |
-| `token-url` | — | URL on the host backend returning `{ "token": "…" }` for the signed-in user. See [Identity](/docs/identity) |
+| `token-url` | — | URL on the host backend returning `{ "token": "…" }` for the signed-in user. Resolved relative to the page. See [Identity](/docs/identity) |
+| `require-identity` | off | Never fall back to an anonymous visitor when identity cannot be obtained |
 | `endpoint` | the directory `widget.js` was served from | Base URL of the manager API. Only needed when the script and the API live at different URLs |
 
-The element also exposes two members for host apps:
+The element also exposes these hooks for host apps:
 
 - `element.tokenProvider = async () => token` — for a SPA holding its token in memory instead of exposing `token-url`.
 - `element.logout()` — clears the stored identity and thread when the host signs a user out.
+- `agent-chat:identity-error` event — raised when a configured `token-url` cannot be used.
 
 With neither configured, the widget asks the manager for a visitor pass, so a
 product with no login still gets one isolated conversation per browser.
--- a/docs/widget.mdx
+++ b/docs/widget.mdx
@@ -68,6 +68,7 @@ Renders as a fixed panel directly in the page — no launcher button.
 The element also exposes these hooks for host apps:
 
 - `element.tokenProvider = async () => token` — for a SPA holding its token in memory instead of exposing `token-url`.
+- `element.refreshIdentity()` — re-resolves the caller after a SPA signs a user in or switches user.
 - `element.logout()` — clears the stored identity and thread when the host signs a user out.
 - `agent-chat:identity-error` event — raised when a configured `token-url` cannot be used.
 
--- a/src/agent_manager/api/deps.py
+++ b/src/agent_manager/api/deps.py
@@ -37,19 +37,14 @@ def get_caller_identity(request: Request) -> CallerIdentity:
 def get_principal(request: Request) -> Principal:
     """The proven caller every conversation route authorizes against.
 
-    The host's session cookie where a deployment names one, otherwise a bearer
-    token. Trusting that cookie is safe because it only reaches us from the
+    A bearer token where the caller supplied one, otherwise the host's session
+    cookie. Trusting that cookie is safe because it only reaches us from the
     host's own origin: cross-site requests cannot read a JSON response, and the
     widget's `application/json` writes are preflighted against a CORS allowlist
     that denies by default.
     """
     identity = get_caller_identity(request)
-    # Where a deployment names a session cookie, that cookie is the host's real
-    # identity and wins. A bearer token there is at most a visitor pass the
-    # widget kept from before the user signed in, and letting it take
-    # precedence would hold someone anonymous for as long as the pass lived —
-    # across reloads, because nothing on the client knows the cookie appeared.
-    token = _cookie_token(request, identity.cookie_name) or _bearer_token(request)
+    token = _select_token(request, identity)
     if token is None:
         raise HTTPException(status_code=401, detail=UNAUTHENTICATED_DETAIL)
     try:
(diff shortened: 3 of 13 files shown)


Asaf-prog:
I’m not approving this yet because there are still a few cases where the new identity behavior can be misleading or incorrect.

1. `requireIdentity: false` can behave as `true`.

For example:

```js
window.agentChatConfig = {
  tokenUrl: "/agent-chat/token",
  requireIdentity: false,
};
```

The generic config mapper creates:

```html
<agent-chat
  token-url="/agent-chat/token"
  require-identity="false">
</agent-chat>
```

But the parser uses:

```ts
element.hasAttribute("require-identity")
```

So the result is:

```ts
requireIdentity === true
```

even though the host explicitly configured `false`.

Another example:

```js
window.agentChatConfig = {
  requireIdentity: false
};
```

A login-less product expecting anonymous fallback could suddenly become fail-closed simply because the attribute exists.

This should be covered by tests for both:

```ts
requireIdentity: true
requireIdentity: false
```

2. `fellBackToAnonymous` does not necessarily describe what actually happened.

Currently it is calculated roughly as:

```ts
fellBackToAnonymous = !requireIdentity
```

But that only tells us that anonymous fallback is allowed, not that it actually succeeded.

Example:

```text
GET /agent-chat/token
→ 500
```

The widget emits:

```json
{
  "reason": "unreachable",
  "fellBackToAnonymous": true
}
```

Then:

```text
POST /auth/anonymous
→ 500
```

No anonymous identity was established, but the event already said that we fell back to anonymous.

Another possible case is:

```text
token-url fails
→ event says fellBackToAnonymous=true
→ following request is authenticated by the host session cookie
```

Again, the reported outcome and the real authentication outcome do not match.

I think this field should either describe capability:

```ts
anonymousFallbackEnabled
```

or be emitted only after the fallback outcome is known.

3. The `TokenSource` constructor change may break existing consumers.

Previously:

```ts
new TokenSource(
  "https://api.example",
  "/agent-chat/token",
  provider,
  storage
);
```

Now:

```ts
new TokenSource("https://api.example", {
  tokenUrl: "/agent-chat/token",
  provider,
  storage,
});
```

The new API is cleaner, but `TokenSource` is exported publicly.

So an existing consumer could still have:

```ts
new TokenSource("https://api.example", "/agent-chat/token");
```

After upgrading, that code may no longer configure the token URL correctly.

If `TokenSource` is considered public API, I’d either keep a compatibility overload, for example:

```ts
new TokenSource(endpoint, options)
new TokenSource(endpoint, tokenUrl, provider, storage)
```

or explicitly document this as a breaking change.

The overall direction of the PR is good, the identity failure classification makes sense, and the tests/CI are green. I just don’t think we should approve until the first two behavioral issues are fixed and the exported API compatibility question is intentionally resolved.

AmitAvital1:
All three are fair. Fixed the two behavioural ones in 3b3cda2; the third is a deliberate call I'd rather state than paper over.

**1. `require-identity="false"` resolved to `true`** — confirmed, and it fails in the worst direction: a login-less product explicitly opting out of fail-closed *became* fail-closed, which breaks exactly the zero-config case this PR claimed to preserve. Presence still means `true`, an explicit `"false"` now stays `false`. Tested in all four states (absent / bare / `"true"` / `"false"`) and end-to-end through `applyConfigAttributes` itself, since that mapper is what produces the stringified value. Verified the new case fails on the old parser with your exact symptom.

**2. `fellBackToAnonymous` claimed an outcome it couldn't know** — also right, and both your counterexamples hold: the anonymous mint can itself fail, and a host session cookie may authenticate the next request instead. Renamed to `anonymousFallbackEnabled`, which is what the value actually is.

I went with the rename rather than emitting after the outcome, deliberately: the warning's whole value is being immediate when someone's `token-url` is wrong, and deferring it until the fallback resolves pushes it past the point where it's useful for that. Happy to revisit if you'd rather have the outcome and accept the delay.

**3. `TokenSource` constructor** — leaving it breaking, intentionally, and I'll say so rather than add a shim. It's an internal primitive that's exported mainly so the test suite can drive it; the supported integration surface is `<agent-chat>` and its attributes. A compatibility overload for a signature nobody is plausibly importing is complexity we'd carry indefinitely, and the project is explicitly still pre-1.0 without back-compat guarantees. Noted as a breaking change in the PR description instead — but if you consider the widget's module exports a public contract, say so and I'll add the overload.

AmitAvital1:
Following up on point 1 — your framing pushed me further than my first fix went, and you were right to put it the way you did.

> ב-HTML, attribute בוליאני נחשב פעיל רק בגלל שהוא קיים

That's the actual HTML rule, and it means my parser change was treating the symptom. `<input disabled="false">` is disabled, so strictly speaking `require-identity="false"` → `true` was *spec-correct*. The real defect was upstream: `applyConfigAttributes` stringifying a JS `false` into an attribute at all, producing markup that reads as opting out while HTML says it's on.

Fixed properly in b312d67 — the mapper now removes the attribute for `false` and writes bare presence for `true`, which is how a boolean attribute is meant to be expressed:

```js
applyConfigAttributes(el, { requireIdentity: false });  // → no attribute at all
applyConfigAttributes(el, { requireIdentity: true });   // → require-identity=""
```

I kept the parser lenient about an explicit `"false"` as well. That part *is* a deliberate deviation from HTML semantics: a human hand-writing `require-identity="false"` plainly means off, and honouring that costs nothing now that we never generate it ourselves. Documented as such in the code rather than left implicit.

Tests cover both directions through the mapper and the parser.

AmitAvital1:
Ready for another look. CI green. Since the last review this branch grew one more fix that wasn't part of the original scope — flagging it explicitly rather than burying it in the diff.

**Review feedback (bdd5a4c → b312d67)**
- `require-identity="false"` resolving to `true` — fixed in the parser (3b3cda2), then properly at the source (b312d67): `applyConfigAttributes` no longer stringifies a JS `false` into an attribute at all. Your point about HTML boolean semantics is what pushed it to the right layer.
- `fellBackToAnonymous` → `anonymousFallbackEnabled`, since the value knows permission, not outcome.
- `TokenSource` constructor: still breaking, deliberately, documented rather than shimmed. Open to reversing if you consider the widget's module exports a public contract — that call is yours.

**New in bf74c5d — a bug found after the review, and a design conflict it exposed**

Reported symptom: after signing in, the chat still shows the signed-out state until a full page reload. Cause is the token being cached for the lifetime of the `TokenSource`, so a SPA that logs in without navigating keeps sending the anonymous pass.

The worse part: the fix I had *documented* for this was "call `logout()`" — and `logout()` deletes the visitor pass, which is exactly what the sign-in merge hands over. Anyone following those docs would have silently discarded their pre-login conversations instead of merging them. Two features from #94 cancelling each other out.

Added `refreshIdentity()`, which drops the token in hand but keeps the pass and the open thread, so the merge still runs and the user carries on in the same conversation. `logout()` keeps its meaning. Both halves are pinned by tests, and the merge one fails if `reset()` is ever made to clear the pass.

**Verified**: 662 Python tests, widget unit suite, 24/24 Playwright. Each new test confirmed to fail when its assertion is inverted or the fix reverted — including reproducing your `require-identity="false"` case against the old parser first.

**Still not verified**: the full browser path with a real signed-in session against a live host. Stated plainly because an earlier claim of that kind in #94 turned out to be wrong; I've corrected that PR's description too.

Asaf-prog:
Thanks — the two issues from my previous review are fixed correctly now:

* `requireIdentity: false` no longer becomes `true`.
* `fellBackToAnonymous` was correctly changed to `anonymousFallbackEnabled`, which describes capability rather than claiming the fallback actually succeeded.

I still see one identity issue around the new `refreshIdentity()` flow that I think should be fixed before approval.

`refreshIdentity()` currently does:

```ts
refreshIdentity(): void {
  this.tokens?.reset();
  this.instanceKey += 1;
  if (this.connected) this.render();
}
```

But `TokenSource` captured `tokenProvider` when it was created:

```ts
this.tokens = new TokenSource(this.config.endpoint, {
  tokenUrl: this.config.tokenUrl,
  provider: this.tokenProvider,
  ...
});
```

So this SPA flow can still fail:

```ts
const chat = document.querySelector("agent-chat");

// Initially signed out.
chat.tokenProvider = null;

// Later the user signs in without a page reload.
chat.tokenProvider = async () => "new-user-token";
chat.refreshIdentity();
```

`refreshIdentity()` only resets the cached token; it does not rebuild/reconfigure the `TokenSource`.

That means the existing `TokenSource` may still hold the old provider (`null`) instead of the newly assigned provider.

There is also a race around an identity resolution that is already in flight.

Example:

```text
1. TokenSource starts resolving the old/signed-out identity.
2. The user signs in.
3. refreshIdentity() clears cached.
4. The old request is still pending.
5. The old request completes and can write the old identity back into cached.
```

So a request started before `refreshIdentity()` should not be allowed to restore stale identity state afterwards.

I’d like to see regression tests for both cases:

```text
tokenProvider changes
→ refreshIdentity()
→ next request uses the new provider/token
```

and:

```text
old identity resolution is pending
→ refreshIdentity()
→ old resolution completes
→ it must not overwrite the refreshed identity
```

After those are handled and CI remains green, I’d be comfortable approving this.

AmitAvital1:
Both confirmed and fixed in 2e65c2b.

**1. Stale `tokenProvider`.** `TokenSource` only reads `provider` once, at construction — `refreshIdentity()` was only clearing the cached token, so a provider assigned after the element connected (the exact flow this method exists for) was never seen; the old, closed-over `null`/stale provider kept answering. Fixed by having `refreshIdentity()` rebuild `TokenSource` from scratch via `configure()`, the same path `attributeChangedCallback` already uses.

**2. The race.** Rebuilding via `configure()` happens to close this for `refreshIdentity()` specifically — a fresh object means the old one's eventual resolution writes into memory nobody reads. But `reset()`/`forget()` are also called directly on the existing `TokenSource` (from `logout()`, and by anyone calling `reset()` without going through the element), so I fixed it at the source instead: each resolution now tags itself with a generation number when it starts, `reset()` bumps the generation and disowns the in-flight promise, and a resolution landing under a stale generation is dropped rather than cached. Covers `logout()`'s identical race too, which your report didn't touch but shares the same root cause.

Added your two scenarios as tests, and — since this branch already had one thing verified wrong the hard way — confirmed each one is a real regression: reverted the corresponding fix, watched the new test fail with the exact symptom you described, restored it, watched it pass. Full gate after: 663 Python tests, widget unit + typecheck, 24/24 Playwright.

Asaf-prog:
Thanks — the two issues from my previous review are fixed correctly now:

* `refreshIdentity()` rebuilds `TokenSource`, so a `tokenProvider` assigned or changed after the widget connects is picked up.
* The stale in-flight resolution race is handled with the generation guard, and the new regression tests cover both cases.

I still see one auth semantic issue before approval.

Right now the backend does:

```python
token = _cookie_token(request, identity.cookie_name) or _bearer_token(request)
```

That makes the session cookie win over **every** Bearer token whenever both are present.

The visitor-pass case makes sense:

```text
visitor Bearer + authenticated host cookie
→ host cookie should win
```

But an authenticated Bearer token is different.

For example:

```text
Cookie:
user = asaf

Authorization:
Bearer <valid host token for noam>
```

With the current code, the request runs as `asaf` because the cookie always wins, even though the caller explicitly supplied an authenticated Bearer token for `noam`.

I think the intended precedence should be closer to:

```text
authenticated host Bearer
→ Bearer wins

anonymous visitor Bearer + host cookie
→ cookie wins

cookie only
→ cookie wins

visitor Bearer only
→ visitor identity
```

So I would avoid changing the general precedence from Bearer-first to cookie-first. Instead, resolve whether the Bearer is an anonymous visitor pass and only let the cookie override that specific case.

One more integration item: #96 has now landed on `main` and renamed the public auth configuration from `AGENT_AUTH_*` to `EXTRA_AUTH_*`.

This branch still contains the old names in its docs, so please update/rebase against the latest `main`, preserve the `EXTRA_AUTH_*` naming, and rerun CI.

Once those two things are addressed, I’d be comfortable approving this.

AmitAvital1:
Both addressed in 3a49342, rebased onto latest `main`.

**Bearer precedence.** You're right, and the diagnosis is sharper than mine was: I fixed a visitor-pass problem with a rule about *where a token arrived*, when the thing that actually matters is *what the token is*. `cookie or bearer` quietly made it impossible to call as anyone but the browser's current session — a backend acting on a user's behalf, or a page holding a freshly minted token, both silently ran as whoever the cookie named.

Precedence now asks the bearer what it claims to be:

```python
bearer = _bearer_token(request)
cookie = _cookie_token(request, identity.cookie_name)
if bearer is None or (cookie is not None and not identity.resolver.names_a_host_user(bearer)):
    return cookie
return bearer
```

`names_a_host_user()` reads the `kid` header only — the same unverified header routing `resolve()` already does to pick a source. That is deliberate and I'd rather say it out loud than have you find it: it decides *which of two tokens to use*, never what anyone is allowed to do. Whichever wins still goes through `resolve()`, and the signature check there is the only thing that grants anything. Forging a `kid` gets you as far as having your unsigned token rejected instead of your cookie honoured.

A bearer too malformed to read is not counted as a claim, so cookie + garbage bearer keeps behaving as it does today rather than turning into a new 401.

All four rows of your table hold, and the third and fourth are already pinned by existing tests. Added the first as `test_a_host_bearer_token_outranks_the_session_cookie` — literally your `asaf` cookie / `noam` bearer case. Confirmed it is a real regression before trusting it: reverted to `cookie or bearer`, and it failed with `assert [] == ['0f3ec23…']` — the conversation filed under Asaf, exactly the symptom you described — then restored and it passes. `test_a_visitor_pass_does_not_shadow_the_host_session_cookie` still passes unchanged, so the case that started all this is not regressed.

**#96 / `EXTRA_AUTH_*`.** Rebased onto `main`, no conflicts. This branch's docs carry `EXTRA_AUTH_*` only. Also renamed 13 `agent_auth_*` kwargs in the auth tests — they were still on the legacy path and each one was emitting a deprecation warning through the suite.

CI green on the rebase: `check` and `ui` both pass.

Asaf-prog (approved):

PR #102, opened 2026-08-15
Feat/shared tool usage

Asaf-prog:
## Shared tool usage across agents

Tool usage was collected in a Python list seeded from `GraphState["used_tools"]`,
merged upward by `OrchestratorNode` after each child returned, and handed back as a
state delta. That ownership model caused three problems:

- an agent could not see what another agent had already run, because the merge
  happened only *after* the child finished;
- nothing survived the run, so a later turn of the same conversation started blind;
- `AgentNode` owned persistence, approval, execution, hooks and tracking at once,
  and a distributed backend had nowhere to plug in.

### Approach

Execution metadata now has its own subsystem, `agent_engine/tool_usage/`, with the
repository as the source of truth. `GraphState["used_tools"]` is removed entirely.

```
Tool execution
      │
      ▼
ToolUsageTracker ──► ToolUsageRepository        (conversation → run → agent → tool call)
                            │
              ┌─────────────┴─────────────┐
              ▼                           ▼
   ToolUsageContextProvider        as_usage_records
   private model context           RunResult.used_tools
```

- **`ToolUsageRepository`** (Protocol, two read scopes + one write) persists one
  record per logical invocation, identified by `(run_id, tool_call_id)`. `record`
  is an upsert on that identity, so an approval resume updates rather than
  duplicates. Ships with a process-local adapter; the contract suite also runs
  against a SQLite adapter so a future Redis/PostgreSQL adapter has a behavioural
  target to meet.
- **`ToolUsageTracker`** writes outcomes (`succeeded` / `failed` / `denied`) and
  decides nothing else.
- **`ToolUsageContextProvider`** projects records into a private system-role block,
  re-read before *every* model turn, so a synthesis turn reflects what the children
  it just called actually did.
- **`ToolInvoker`** takes over the tool pipeline — identity, limits, approval gate,
  idempotency, provider execution, hooks, recording — leaving nodes with prompt
  building and the model loop. It is a move, not a second execution path.

Both leaf agents and orchestrators receive the context. Backend selection lives only
in `agent_manager/composition.py`.

### Design decisions worth reviewing

**Conversation *and* run scope.** Records carry both. A follow-up user message is a
new `run_id`, so conversation scope is what gives continuity across turns; run scope
keeps execution-level traceability and remains what `RunResult.used_tools` reports.

**The projection is not the record.** Only agent, tool name and status reach the
model — no arguments, results, ids, timestamps or error text. A trimmed record says
so rather than passing itself off as complete.

**Non-duplication.** An invocation the asking agent made in the run it is currently
executing is skipped while its own `AIMessage` + `ToolMessage` pair is in the turn —
the model already has it in full. The same agent's calls from an earlier run or node
entry *are* shown.

**`list_executed_tools`.** An orchestrator's children are bound to it as tools, so
asked "which tools have you run?" it named the child agent it had delegated to — and
kept doing so through four rounds of instruction and record wording, because from its
own tool list that answer is true. Orchestrators are now bound a read-only engine tool
that returns the executed tools from the repository, so the answer comes from data
rather than from an instruction the model can ignore. It is not approval-gated
(orchestrator-level tools never were), not counted against child-agent limits, and
not itself recorded. A configured child of that name wins.

**Failure semantics.** Recording is observability: a write failure is logged and
swallowed, so a metadata problem cannot fail a completed tool call. A *read* failure
on the explicit report is **not** swallowed — reporting "nothing has run" when the
record is unreadable would invite repeating an action that already took effect.

### Contract changes (ADR 0001)

- `GraphState["used_tools"]` removed. `RunResult.used_tools` keeps its shape for
  callers and is now projected from the repository.
- `ToolUsageStatus` gains `"denied"`. The HTTP/widget field is a free string.
- `LangGraphEngine.__init__` gains `tool_usage_repository` (defaults to in-memory).
- Orchestrators are bound `list_executed_tools`; child agents are described to their
  parent as delegations rather than actions.
- Two identical calls by one agent in one run now collapse to a single record — they
  are one logical invocation. The previous list appended twice.

### Human-in-the-Loop

Unchanged and verified: approve, deny, approve-for-this-session, suspension and
resume all behave as before. One logical invocation leaves one record across a
suspend and its replay, with the same `run_id`, agent and `tool_call_id` as the
`ApprovalRecord`. Approval decides *whether* a call runs; tracking records *what
happened to it* — separate subsystems sharing identifiers, not responsibilities.

### Testing

765 tests pass (`make check` clean). New coverage: repository contract against two
independent backends, cross-agent and cross-turn awareness, per-agent attribution,
run/conversation isolation, context freshness between model turns, delegation never
offered as a tool, the reporting tool's content and failure mode, and the HITM
lifecycle. Also pinned: the record never becomes conversation history, never reaches
the answer, and never repeats what the current turn already shows the model.

### DIFF
--- a/src/agent_engine/engine/langgraph/executed_tools_tool.py
+++ b/src/agent_engine/engine/langgraph/executed_tools_tool.py
@@ -1,0 +1,66 @@
+"""The engine's own read-only tool for reporting what has already executed.
+
+An orchestrator's children are bound to it as tools, so asked "which tools have
+run?" a model answers from that list and names a child agent — a claim no system
+instruction reliably corrected. Giving it a tool turns the question into a tool
+call whose result comes from the usage repository, which the model cannot
+contradict from its own bindings.
+
+The tool performs no side effect: it reads execution metadata and returns text.
+"""
+
+from __future__ import annotations
+
+import logging
+
+from langchain_core.tools import BaseTool, StructuredTool
+
+from agent_engine.engine.langgraph.helpers import current_usage_scope
+from agent_engine.logging_config import log
+from agent_engine.tool_usage.context import (
+    EXECUTION_RECORD_UNAVAILABLE,
+    ToolUsageContextProvider,
+    format_executed_tools,
+)
+
+logger = logging.getLogger(__name__)
+
+EXECUTED_TOOLS_TOOL_NAME = "list_executed_tools"
+
+_DESCRIPTION = (
+    "List the tools that have actually been executed so far in this conversation, "
+    "with the agent that ran each one and its outcome. Call this whenever the user "
+    "asks which tools have run, what has been done, or whether an action already "
+    "happened. The result is authoritative execution metadata: answer from it, not "
+    "from the agents in your own tool list."
+)
+
+
+def build_executed_tools_tool(provider: ToolUsageContextProvider, agent_id: str) -> BaseTool:
+    """Bind the usage provider into a no-argument reporting tool."""
+
+    async def list_executed_tools() -> str:
+        scope = current_usage_scope(agent_id)
+        try:
+            context = await provider.get_executed_tools(scope)
+        except Exception as exc:
+            # Never raise out of a tool the model called, and never let a read
+            # failure be reported as "nothing has run" — that would invite a
+            # repeat of an action that already took effect.
+            log(
+                logger,
+                logging.WARNING,
+                "execution record unavailable",
+                agent=agent_id,
+                run_id=scope.run_id,
+                conversation_id=scope.conversation_id,
+                error=type(exc).__name__,
+            )
+            return EXECUTION_RECORD_UNAVAILABLE
+        return format_executed_tools(context)
+
+    return StructuredTool.from_function(
+        coroutine=list_executed_tools,
+        name=EXECUTED_TOOLS_TOOL_NAME,
+        description=_DESCRIPTION,
+    )
--- a/src/agent_engine/engine/langgraph/tool_invoker.py
+++ b/src/agent_engine/engine/langgraph/tool_invoker.py
@@ -1,0 +1,431 @@
+"""Coordination of a single tool invocation for one agent.
+
+``ToolInvoker`` owns the path from "the model asked for this tool" to "here is
+the text the model gets back": identity, execution limits, the Human-in-the-Loop
+gate, idempotency, provider execution with lifecycle hooks, and handing the
+outcome to the usage tracker. There is exactly one such path — the approval flow
+and the tracking flow are steps in it, never a second pipeline.
+
+Keeping it out of the node leaves ``AgentNode`` with prompt building and the
+model loop, and lets the invoker be constructed and tested on its own.
+"""
+
+from __future__ import annotations
+
+import logging
+import time
+from collections.abc import Mapping
+from dataclasses import dataclass, field
+from typing import Any
+
+from langchain_core.tools import BaseTool
+
+from agent_engine.approvals.coordinator import ApprovalCoordinator
+from agent_engine.approvals.invocation import ToolInvocation
+from agent_engine.approvals.manager import ToolExecutionManager, execution_id_for
+from agent_engine.core.spec import AgentSpec
+from agent_engine.logging_config import log
+from agent_engine.runtime.execution import (
+    ExecutionLimitExceeded,
+    blocked_message,
+    current_execution,
+    log_limit,
+)
+from agent_engine.runtime.hooks import (
+    HookManager,
+    ToolCallContext,
+    ToolRequestContext,
+    ToolResultContext,
+    current_run_context,
+)
+from agent_engine.runtime.hooks.models import ToolStatus
+from agent_engine.runtime.tool_models import ToolProviderName
+from agent_engine.tool_usage.models import ToolCallIdentity, stable_tool_call_id
+from agent_engine.tool_usage.tracker import ToolUsageTracker
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True)
+class ExecuteTool:
+    """Approval-gate result: the gate is open — the caller may run the tool."""
+
+
+@dataclass(frozen=True)
+class DenyTool:
+    """Approval-gate result: the gate is closed — do not run the tool.
+
+    ``message`` is the structured, model-facing result to return in place of
+    execution (a normal denial, not a system error).
+    """
+
+    message: str
+
+
+ToolGate = ExecuteTool | DenyTool
+
+
+@dataclass(frozen=True)
+class AgentToolBinding:
+    """The tools one agent may call, plus where each of them came from."""
+
+    tools: dict[str, BaseTool] = field(default_factory=dict)
+    mcp_tool_names: frozenset[str] = frozenset()
+    mcp_server_by_tool: Mapping[str, str] = field(default_factory=dict)
+
+    def get(self, name: str) -> BaseTool | None:
+        return self.tools.get(name)
+
+    def provider_of(self, name: str) -> ToolProviderName:
+        return "mcp" if name in self.mcp_tool_names else "local"
+
+    def server_of(self, name: str) -> str | None:
+        return self.mcp_server_by_tool.get(name)
+
+
+@dataclass(frozen=True)
+class _ToolCall:
+    """One resolved, about-to-run tool call: the tool plus its stable identity.
+
+    Bundling these fields once removes the repeated ``agent/tool/provider/server``
+    argument lists that the limit, gate, logging, hook, tracking, and execution
+    steps would each otherwise rebuild from the raw ``tc`` dict.
+    """
+
+    tool: BaseTool
+    identity: ToolCallIdentity
+    args: dict[str, Any]
+    exec_id: str
+
+    @property
+    def name(self) -> str:
+        return self.identity.tool_name
+
+    @property
+    def provider(self) -> ToolProviderName:
+        return self.identity.provider
+
+    @property
+    def server_id(self) -> str | None:
+        return self.identity.server_id
+
+    @property
+    def tool_call_id(self) -> str:
+        return self.identity.tool_call_id
+
+    @property
+    def run_id(self) -> str:
+        return self.identity.run_id
+
+    @property
+    def description(self) -> str:
+        return self.tool.description
+
+
+def _elapsed_ms(start: float) -> int:
+    """Whole milliseconds elapsed since a ``time.perf_counter()`` reading."""
+    return int((time.perf_counter() - start) * 1000)
+
+
+class ToolInvoker:
+    """Runs one agent's tool calls: gate, execute, record.
+
+    All collaborators are injected, and none of them is per-run state, so one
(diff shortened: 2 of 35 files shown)


AmitAvital1 (commented):
Ran both halves of the project on the branch, not just `make check`: `typecheck:widget`, `test:widget`, `typecheck:e2e` all pass, and I rebuilt `widget.js` with esbuild — byte-identical to the committed bundle, so it isn't stale. 764 python tests pass; the one failure (`test_run_persistence.py::test_agentctl_run_without_session_prints_reusable_generated_session`, `stderr not separately captured`) reproduces on `main` and is a local click-version issue, not yours.

Coverage on the new subsystem is genuinely high — `tool_usage/` is 100% across all six modules, `tool_invoker.py` 96%. `stable_tool_call_id` being derived from the call rather than the provider's message id, and the write/read failure asymmetry, are both the right calls and well argued in the docstrings.

Four things below. Only the first is one I'd want fixed before merge.

---

**1. `_is_visible` suppresses a record the current turn cannot see** — `tool_usage/context.py:281-289`

The filter has only the tool *name*, so it cannot tell "the call this turn made" from "the same agent's call in an earlier node entry of the same run" — which is exactly what the docstring on line 275 promises it can. Probe: two records, same tool/agent/run, different `tool_call_id`, `already_visible={"all_user_details"}` → `entries=()` and `render()` returns `None`, so the model gets no execution block at all. Control with a different `agent_id` → 1 entry, confirming the key is name+agent+run.

Where it bites: an orchestrator delegating twice to one agent in a single run. The second entry emits its own call, the run-wide duplicate guard in `runtime/execution.py:92` blocks it with "use the previous result", and this filter then removes the one record that proved the tool ran — so the mitigation this PR adds is switched off precisely when it is needed. Cheapest fix is to drop the `already_visible` filter: three fields per entry is a few tokens, and it removes the whole class. Otherwise key it on a per-invocation id rather than a name.

---

**2. The projection is bounded, the read is not** — `tool_usage/repository.py:31-41`

Measured with a counting repository on a 3-tool leaf agent over 10 turns in one conversation: 4 `list_for_conversation` calls per turn, and the records pulled grows linearly — turn 10 loads 117 records to render at most 50. 660 records pulled in total to store 30, a 22× amplification on one agent at one depth. In-memory this is free; it is the hot path for the Redis/Postgres adapter the Protocol exists for. The Protocol is also the hardest thing to change once adapters exist — worth adding a `limit` to both read methods now rather than after.

Non-blocking.

---

**3. The replay-after-resume branch is unreached by all 765 tests** — `engine/langgraph/tool_invoker.py:248-258`

Coverage puts the whole body of `_cached_result` past the early return in the missing list. Every HITL test suspends on the only tool call, so there is never a prior successful execution to hit the cache. I built the missing case — two tools, allow-for-session on the first, second approval triggers the replay — and it behaves correctly: the first tool ran once, two records. So not a bug, but this is the branch that stops a completed side effect from happening twice, and it would be good to have that test in the suite rather than in a reviewer's scratchpad.

Non-blocking.

---

**4. Does `list_executed_tools` earn its permanent surface?**

I believe the evidence that four rounds of prompt wording failed. But `### Tools executed` in the context block already carries the same data from the same records, so this is a second model-facing vocabulary for it plus a tool bound to every orchestrator forever, for a question users ask rarely. Genuinely a judgement call — I'd just like to hear the cost side of it weighed.

---

**On the release marker:** commits are properly conventional, but none carries `!` or a `BREAKING CHANGE:` footer, and `GraphState["used_tools"]` was removed — the three `examples/` scripts and `agentctl/chat.py` in this diff are the tell. release-please will cut a minor from this as it stands. (Nice side effect in `agentctl/chat.py`: the duplicated `_ApprovalEngine` Protocol replaced by importing the real one.)

**Overlap with #103:** that PR scopes the duplicate-tool-call guard to one agent invocation; `runtime/execution.py` is untouched here, and `ToolInvoker.invoke` keeps `_enforce_limits` (176) ahead of `_cached_result` (185), so the false block survives this refactor. Landing #103 first is 4 files and avoids the fix being buried under this move; conflicts are limited to `nodes.py` and `helpers.py`.

src/agent_engine/tool_usage/context.py
```diff
@@ -0,0 +1,307 @@
...
+
+        Only the asker's own calls, in the run it is currently executing, can be
+        in its message list — another agent's call, or the same agent's call in
+        an earlier run or an earlier node entry, is not.
+        """
+        call = record.call
+        return (
+            call.tool_name in already_visible
+            and call.agent_id == scope.agent_id
+            and call.run_id == scope.run_id
```
AmitAvital1: This suppresses a record the current turn cannot see. With only the tool *name*, the filter can't tell "the call this turn made" from "the same agent's call in an earlier node entry of the same run" — which line 298 promises it can.

Probe: two records, same tool/agent/run, different `tool_call_id`, `already_visible={"all_user_details"}` → `entries=()` and `render()` returns `None`, so the model gets no execution block at all. Control with a different `agent_id` → 1 entry, confirming the key is name+agent+run.

Where it bites: an orchestrator delegating twice to one agent in one run. The second entry emits its own call, the run-wide guard at `runtime/execution.py:92` blocks it with "use the previous result", and this filter then removes the one record proving the tool ran — so the mitigation this PR adds is off exactly when it's needed.

Cheapest fix is to drop `already_visible` entirely: three fields per entry is a few tokens and it removes the class. Otherwise key it on a per-invocation id, not a name. **This is the one I'd fix before merge.**

src/agent_engine/tool_usage/repository.py
```diff
@@ -0,0 +1,43 @@
...
+
+    async def record(self, record: ToolInvocationRecord) -> None:
+        """Store the outcome of one logical tool invocation."""
+        ...
+
+    async def list_for_run(self, run_id: str) -> Sequence[ToolInvocationRecord]:
+        """Return this run's records in call order (empty when unknown)."""
+        ...
+
+    async def list_for_conversation(self, conversation_id: str) -> Sequence[ToolInvocationRecord]:
```
AmitAvital1: The projection is bounded to 50, this read isn't. Measured with a counting repository, 3-tool leaf agent, 10 turns in one conversation: 4 calls to this method per turn, and turn 10 pulls 117 records to render at most 50 — 660 pulled in total to store 30, a 22× amplification on one agent at one depth.

Free in-memory; it's the hot path for the Redis/Postgres adapter this Protocol exists for. And the Protocol is the hardest piece to change once adapters exist — worth a `limit` on both read methods now rather than after. Non-blocking.

src/agent_engine/engine/langgraph/tool_invoker.py
```diff
@@ -0,0 +1,426 @@
...
+            logger,
+            logging.INFO,
+            "duplicate tool execution prevented",
+            agent=self._spec.id,
+            tool=call.name,
+            tool_call_id=call.tool_call_id,
+            execution_id=call.exec_id,
+        )
+        await self._usage.record_success(call.identity)
+        return cached.result
```
AmitAvital1, line 262: Coverage puts this whole block in the missing list — unreached by all 765 tests. Every HITL test suspends on the only tool call, so there's never a prior successful execution to hit the cache.

I built the missing case (two tools, allow-for-session on the first, second approval triggers the replay) and it behaves correctly: first tool ran once, two records. So not a bug — but this is the branch that stops a completed side effect from running twice, and I'd rather that test lived here than in my scratchpad. Non-blocking.

src/agent_engine/engine/langgraph/tool_invoker.py
```diff
@@ -0,0 +1,426 @@
...
+        blocked = self._enforce_limits(call)
+        if blocked is not None:
+            return blocked
+
+        gate = await self._gate_tool_call(call)
+        if isinstance(gate, DenyTool):
+            await self._usage.record_denied(call.identity)
+            return gate.message
+
+        cached = await self._cached_result(call)
```
AmitAvital1, line 189: Worth knowing while this moves: `_enforce_limits` ahead of `_cached_result` means the run-wide duplicate guard shadows the cache for any repeat inside one run scope. `tool_call_id` is derived from `[run_id, node_path, provider, server_id, name, args]`, so the cache *would* hit — the block just gets there first and returns "use the previous result" to an agent that can't see it.

#103 scopes that guard to one agent invocation, which routes this straight into your cache. `runtime/execution.py` is untouched here, so the behaviour survives the refactor; landing #103 first keeps the fix from being buried under the move.

src/agent_engine/engine/langgraph/executed_tools_tool.py
```diff
@@ -0,0 +1,66 @@
...
+from agent_engine.logging_config import log
+from agent_engine.tool_usage.context import (
+    EXECUTION_RECORD_UNAVAILABLE,
+    ToolUsageContextProvider,
+    format_executed_tools,
+)
+
+logger = logging.getLogger(__name__)
+
+EXECUTED_TOOLS_TOOL_NAME = "list_executed_tools"
```
AmitAvital1, line 28: I believe the evidence that four rounds of prompt wording failed. But `### Tools executed` in the context block already carries the same data from the same records, so this is a second model-facing vocabulary for it plus a tool bound to every orchestrator forever, for a question users ask rarely.

Genuinely a judgement call, not a defect — I'd just like the cost side weighed out loud. Non-blocking.

AmitAvital1 (commented):
Follow-up to my summary comment — anchoring each point to the line it concerns, since three of them are edits rather than discussion.

Recap of what I ran: `typecheck:widget`, `test:widget`, `typecheck:e2e` all pass, and `widget.js` rebuilds byte-identical to the committed bundle. The one failing python test reproduces on `main` (local click version), not yours.

Asaf-prog:
Thanks for the review. I addressed all unresolved comments and rebased `feat/shared-tool-usage` onto the latest `main`.

Changes made:

- Removed the incorrect `already_visible` filtering. The repository remains the source of truth, and repeated calls to the same tool or child agent remain visible as separate logical invocations.
- Added bounded repository reads with optional invocation-kind filtering. Reads return the newest matching records in chronological order, and the context provider uses `limit + 1` to detect truncation without loading the full history.
- Added contract coverage for both the in-memory and SQLite adapters, including ordering, limits, filtering, and avoiding under-filled tool-only results.
- Added HITL replay coverage for chained approvals. When the second approval resumes the graph, the first tool’s cached result is reused, its side effect runs only once, and no duplicate usage record is created.
- Integrated PR #103’s `current_invocation` execution scoping instead of introducing another mechanism. Identical calls remain blocked within one node activation, while graph/HITL replay can reach the idempotency cache.
- Kept `list_executed_tools` for explicit reporting questions because passive context alone did not reliably prevent models from identifying delegated agents as executed tools. It remains read-only, unrecorded, outside tool/child-call limits, uses centralized wording, and yields to a configured child with the same name.
- Updated the ADR and tool/MCP documentation to record these decisions and tradeoffs.

Validation completed successfully:

- `make check` passed
- 782 tests passed
- Formatting, linting, and mypy all passed

Latest commit: `00d587c fix(tool-usage): address review feedback`

AmitAvital1 (approved):

PR #108, opened 2026-08-16
Feat/support cancel interrupt

Asaf-prog:
Cancel / interrupt support and immutable conversation branches
Summary

A conversation is now an immutable message tree with an explicitly selected active branch, and every turn — new, resumed, or suspended on an approval — can be interrupted with an authoritative, persisted outcome.

Previously messages were one append-only list (the schema already carried an unused parent_message_id). Cancelling a run left its user message behind but the history API could not project the run's real state, so the widget erased all evidence of the stopped turn. Editing an earlier message could not safely mutate that row or reuse its descendants, since every later answer was produced from the original content. The browser also received no response until the engine emitted its first event, which made "stop before the first token" depend on cancelling an ordinary request handler instead of a response-owned stream.

Design and rationale: docs/adr/0002-conversation-lifecycle-and-immutable-branches.md.

What changed

Immutable message tree

ConversationMessage.parent_message_id links a message to its predecessor; ConversationSession.head_message_id selects the active root-to-leaf path.
A normal turn appends at the head. An edit appends a new user message at the edited message's parent and atomically selects it as the new head — existing descendants stay stored and unchanged.
A late assistant result is stored on its original branch, and selects itself only if its parent is still the active head.
Model context and the history API traverse only the selected ancestry. No separate branch_id: the leaf message already identifies a branch.

Cancellation lifecycle

The manager prepares and persists the turn before returning the streaming response; the first SSE frame is turn_started, carrying the authoritative run_id and message_id. Graph execution is owned by the response generator, so a disconnect at any point terminalizes the run.
The run is registered before turn_started, so disconnecting before engine iteration can no longer leave a run stuck in running.
A terminal engine event is acknowledged before the run becomes COMPLETED, and the assistant message is persisted before that event reaches HTTP — a disconnect racing acknowledgement cancels the run rather than dropping a queued final answer.
New turns and resumed turns share one engine-owned graph-task cancellation path. Resume keeps the original run_id and moves PENDING_APPROVAL → RESUMING → RUNNING before emitting work.
A suspended approval stays resumable after its SSE response ends; the owner can terminalize it explicitly. Cancellation competes atomically with the approval claim, so exactly one of cancel/resume wins.
The synthetic UI label Generation stopped is never persisted or sent to the engine.

Contract / API

History items expose message_id, run_id, and status.
Send and stream requests accept an optional edit_message_id.
New manager-owned turn_started stream event.
New POST .../runs/{run_id}/approvals/{approval_id}/cancel — cancellation is a lifecycle transition, not an approval decision and not DENY.
New POST .../approvals/{approval_id}/decision/stream; its resume_started event carries the original run_id, and disconnecting it cancels the graph task.
Sessions gain head_message_id; manager persistence gains durable run records via a SQL RunRepository adapter injected into both the engine and the conversation service.
Failed or disconnected stream mutations are never auto-replayed over the non-streaming endpoint — without a client idempotency key, a delivery failure cannot prove the original request changed no state.

Persistence / composition

With no database URL configured the manager composes process-local conversation and run repositories and opens no SQL connection. AGENT_DB_URL or DATABASE_URL opts into SQL persistence and Alembic.
No new migration revision: no database environment exists yet, so the existing fresh-database baseline revision was updated in place.

Widget

Stop control for an in-flight turn, message editing that branches, cancelled turns rendered from persisted run state, and resume controls that stay available while an approval is still active.
Known limits
External side effects completed before cancellation are not rolled back.
If the approval claim wins the race, /cancel returns a conflict.
Token budget stays cumulative across executed branches — editing does not refund model/tool work already consumed.
Usage reported by a provider only after a cancelled request cannot be counted exactly.
Without a configured database URL, history and run state are lost on restart.
Tests
Repository contract tests run ancestry/head behavior against both the memory and SQL adapters.
Service tests assert the exact model context after edit and after cancel.
API and browser (Playwright) tests cover lifecycle projection, edit identity, immutable downstream replacement, cancellation display, and reload.
New coverage in tests/approvals/ for cancel-vs-claim races and streamed resume.

### DIFF
--- a/src/agent_manager/api/routes.py
+++ b/src/agent_manager/api/routes.py
@@ -5,9 +5,9 @@ from __future__ import annotations
 import dataclasses
 import json
 import logging
-from collections.abc import AsyncIterator, Iterator
+from collections.abc import AsyncGenerator, AsyncIterator, Iterator
 from contextlib import contextmanager
-from typing import Annotated, Any
+from typing import Annotated, Any, cast
 
 from fastapi import APIRouter, Depends, HTTPException
 from fastapi.responses import StreamingResponse
@@ -23,6 +23,7 @@ from agent_manager.api.deps import CallerIdentity, get_caller_identity, get_prin
 from agent_manager.api.schemas import (
     AnonymousPassResponse,
     ApprovalDecisionRequest,
+    CancelRunResponse,
     ConversationSummary,
     CreateConversationRequest,
     CreateConversationResponse,
@@ -39,7 +40,9 @@ from agent_manager.api.schemas import (
 from agent_manager.application import (
     ConversationAccessDenied,
     ConversationAlreadyExists,
+    ConversationBranchConflict,
     ConversationLinkRefused,
+    ConversationMessageNotFound,
     ConversationNotFound,
     ConversationService,
     ConversationTokenBudgetExceeded,
@@ -65,6 +68,8 @@ _HTTP_ERRORS: dict[type[Exception], tuple[int, Any]] = {
     ConversationAccessDenied: (403, "conversation owned by another user"),
     ConversationAlreadyExists: (409, "conversation id already taken"),
     ConversationTokenBudgetExceeded: (429, _BUDGET_EXCEEDED_DETAIL),
+    ConversationMessageNotFound: (404, "message not found on active conversation branch"),
+    ConversationBranchConflict: (409, "conversation branch changed; reload and try again"),
     ConversationLinkRefused: (403, "a visitor cannot adopt another visitor"),
 }
 
@@ -134,7 +139,17 @@ async def list_conversations(service: Service, caller: Caller) -> list[Conversat
 async def list_messages(conversation_id: str, service: Service, caller: Caller) -> list[MessageOut]:
     with _as_http_error():
         msgs = await service.history(conversation_id, caller)
-    return [MessageOut(role=m.role, content=m.content, created_at=m.created_at) for m in msgs]
+    return [
+        MessageOut(
+            message_id=m.message_id,
+            run_id=m.run_id,
+            role=m.role,
+            content=m.content,
+            status=m.status,
+            created_at=m.created_at,
+        )
+        for m in msgs
+    ]
 
 
 @router.get("/conversations/{conversation_id}/usage", response_model=TokenBudgetResponse)
@@ -189,7 +204,12 @@ async def send_message(
 ) -> SendMessageResponse:
     try:
         with _as_http_error():
-            result = await service.send(conversation_id, body.message, caller)
+            result = await service.send(
+                conversation_id,
+                body.message,
+                caller,
+                edit_message_id=body.edit_message_id,
+            )
     except HTTPException:
         raise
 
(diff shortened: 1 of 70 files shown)


src/agent_manager/config.py
```diff
@@ -96,7 +96,9 @@ class Settings(BaseSettings):
 
     extra_db_backend: Literal["sqlite", "postgres"] = "sqlite"
     extra_db_url: str | None = None
-    database_url: str = "sqlite+aiosqlite:///chat.db"
+    # Persistence is opt-in. Without either URL the manager uses its process-local
+    # repository adapters and never opens a database connection.
+    database_url: str | None = None
```
AmitAvital1: This flips the default from a persistent sqlite file to process memory. Any deployment relying on the default silently loses its history on the next restart, and nothing warns at startup. Can this be split out of the cancel/interrupt PR?

src/agent_manager/config.py
```diff
@@ -174,5 +176,11 @@ def from_values(cls, **values: Any) -> Settings:
...
-    def effective_database_url(self) -> str:
-        return normalize_database_url(self.extra_db_url or self.database_url, self.extra_db_backend)
+    def effective_database_url(self) -> str | None:
+        url = self.extra_db_url or self.database_url
+        return normalize_database_url(url, self.extra_db_backend) if url else None
+
+    @property
+    def uses_process_memory(self) -> bool:
+        url = self.effective_database_url
+        return url is None or ":memory:" in url
```
AmitAvital1: Substring match — a db path that happens to contain `:memory:` silently drops persistence. Parse the URL instead.

src/agent_manager/application/service.py
```diff
@@ -253,45 +346,157 @@ async def decide_approval(
...
+                run_id,
+                approval_id,
+                decision,
+                caller_user_id=session.user_id,
+                caller_session_id=session.session_id,
+            )
+            try:
+                async for event in engine_stream:
+                    if event.type == "final":
+                        await self._persist_assistant_turn(
```
AmitAvital1: `_persist_stream_final` shields this same write for new turns, but here it's unshielded. An abort landing on this await leaves the run `COMPLETED` with no assistant message — the exact race the PR closes for `stream_turn`.

src/agent_manager/application/service.py
```diff
@@ -322,6 +529,56 @@ async def _persist_assistant_turn(
...
+        self, messages: list[ConversationMessage]
+    ) -> list[ConversationMessage]:
+        completed = {
+            message.run_id
+            for message in messages
+            if message.role == Role.ASSISTANT and message.run_id is not None
+        }
+        statuses: dict[str, str] = {}
+        if self._run_repository is not None:
+            for run_id in {message.run_id for message in messages if message.run_id is not None}:
```
AmitAvital1: One `get()` per run id, so N+1 on every `GET /messages`. A single batched lookup would do.

src/agent_manager/api/routes.py
```diff
@@ -266,27 +374,46 @@ def _to_stream_event(event: RunStreamEvent) -> StreamEventOut:
...
+                message_id=turn.message_id,
+            ).model_dump(exclude_none=True)
+            yield f"event: turn_started\ndata: {json.dumps(started)}\n\n"
             async for event in stream:
+                suspended = suspended or event.type == "pending_approval"
                 payload = _to_stream_event(event).model_dump(exclude_none=True)
                 yield f"event: {event.type}\ndata: {json.dumps(payload)}\n\n"
+            exhausted = True
         except Exception:
+            await service.fail_turn(turn)
```
AmitAvital1, line 405: `exhausted` stays False here, so the `finally` calls `cancel_turn` right after `fail_turn`. That's a no-op today only because `FAILED -> CANCELLED` is rejected. Set `exhausted = True` here too?

src/agent_manager/infrastructure/persistence/sql_repository.py
```diff
@@ -232,11 +264,69 @@ async def append_message_if_absent(
...
+        return _message(row) if row is not None else None
+
     async def list_conversation_messages(
         self, session_id: str, limit: int | None = None
     ) -> list[ConversationMessage]:
         async with self._sessions() as session:
-            rows = await self._message_rows(session, session_id, limit)
+            rows = await self._active_message_rows(session, session_id)
+        if limit is not None:
+            rows = rows[-limit:]
```
AmitAvital1: `limit` used to go to SQL; now every message in the session is loaded and sliced in Python. This is the hot history path.

AmitAvital1 (commented):
Ran `make check` on the branch: format/lint/mypy clean, pytest green (the one `test_run_persistence` failure reproduces on main).

The cancellation + branch core looks right and the tests are good. Main concerns are packaging: the persistence default flip is a breaking change unrelated to cancel/interrupt and should be its own PR.

One more, no diff line to hang it on: `RunLifecycle.succeed(on_completed=...)` has no caller left now that the terminal handshake moved into `StreamChannel` — worth dropping the param and `_noop`.

Asaf-prog:
hanks — all six inline comments and the on_completed note are addressed, in four commits so each concern reviews separately.

Persistence default (0d8b4ae) — agreed, that flip didn't belong here. database_url is back to sqlite+aiosqlite:///chat.db and effective_database_url returns str again, so composition and Alembic behave exactly as before for existing deployments. Docs and the ADR are restored to describing the persistent default. uses_process_memory stays, but now only selects the process-local repositories for an explicitly in-memory URL — that's what tests and throwaway processes use, and it's no longer reachable by default.

:memory: substring match (0d8b4ae) — now parsed with make_url. Empty databases, :memory:, mode=memory, and file: URI filenames are all recognized; non-SQLite backends never are. Regression test: sqlite+aiosqlite:////var/lib/:memory:/chat.db stays persistent.

Unshielded resume write (c312837) — right, that was the same race left open on the resume path. Both terminal paths now go through one _persist_durably helper (the shield extracted out of _persist_stream_final). The test blocks the assistant write, cancels the consumer mid-write, and asserts the message still lands — it fails without the fix. I left decide_approval and complete_turn unshielded on purpose: they're non-streaming, so the answer never becomes observable before the write.

exhausted after fail_turn (c312837) — set. I applied it in ConversationService.stream too, which had the same shape. New test asserts a failed stream stays FAILED and that CANCELLED is never even attempted, so it doesn't silently start passing again if the transition table changes.

N+1 on GET /messages (2600ee2) — RunRepository gains get_many: one IN statement in the SQL adapter, one dict pass in the process-local one. The service makes a single call. Contract tests cover both adapters.

limit sliced in Python (2600ee2) — the SQL adapter now walks parent_message_id from the head in a recursive CTE, with branch.depth < limit - 1 bounding the recursion itself, so a limited read touches only the rows it returns rather than the whole session. get_context_at and the snapshot rebuild share the traversal, and the Python walker is deleted. Traversal stayed a repository concern — nothing SQL-specific moved into the service.

One behavior change worth a look: a non-positive limit now returns [] in both adapters. Previously rows[-0:] quietly returned everything.

on_completed (24a127b) — dropped, along with _noop and the now-unused Callable import.

AmitAvital1 (approved):
All six addressed, and the fixes are better than what I asked for. Re-ran `make check` on 24a127b: format/lint/mypy clean, pytest green (the one `test_run_persistence` failure still reproduces on main).

- `0d8b4ae` — default is back to `chat.db` and `effective_database_url` is `str` again, so nothing changes for existing deployments. `_is_in_memory_sqlite` parsing `mode=memory` and `file:` URI filenames is more than I expected.
- `c312837` — `_persist_durably` is the right shape: one helper, both paths, and the test that cancels the consumer mid-write is the one that matters.
- `2600ee2` — carrying the limit into the recursive CTE means a limited history read touches only the rows it returns. Nice catch on `rows[-0:]`.
- `24a127b` — thanks.

Two nits, neither blocking:
- `alembic.ini:1` still says `AGENT_DB_URL`; everything else moved to `EXTRA_DB_URL`.
- `AgentChatApp.tsx` still has `const canEdit = true` — inline it whenever you next touch that file.

Standing opinion unchanged: this is still a large PR for one branch. Splitting the persistence change out helped; if there's a natural seam between the immutable-branch tree and the cancellation lifecycle, future work in this area would be easier to review split that way.

PR #110, opened 2026-08-20
fix(api): recover the original result on a retried approval decision

### LINKED ISSUE #109: HTTP API doesn't recover from duplicate approval decision
### Problem
The `/approve`, `/reject`, and `/decision` endpoints all trace back to the `_decide` helper function in `src/agent_engine/api/app.py`.  If a client submits a duplicate decision that has already been approved, `engine.resume(...)` raises an `ApprovalAlreadyProcessed` exception, which maps to a `409` status code with no recovery attempt.
```
async def _decide(
    run_id: str,
    approval_id: str,
    decision: ApprovalDecision,
    user_id: str | None,
    session_id: str | None,
) -> InvokeResponse:
    engine = _hitl_engine()
    try:
        result = await engine.resume(
            run_id, approval_id, decision,
            caller_user_id=user_id,
            caller_session_id=_run_context(session_id, run_id=run_id).conversation_id,
        )
    except ApprovalError as exc:
        raise _map_approval_error(exc) from exc  # ← ApprovalAlreadyProcessed → 409, no recovery
    ...
```
In the `agent_manager`, if a duplicate request is made, the `ApprovalAlreadyProcessed` error is caught and `engine.get_processed_result(...)` is called to recover the original result. This is currently not happening in the HTTP layer.

### Impact
There could be a number of reasons why a client ends up making a duplicate decision request (e.g. network timeout, etc). In these situations, receiving a `409` error could be misleading because a client may not know to interpret the `409` error as a successful result. This is also an inconsistency in the codebase, as it is handled differently here than it is in the `agent_manager`. 

### How to Reproduce
1. Start the engine using the bundled starter example
```
agentctl serve --config examples/starter/agents.yaml
```
2. Trigger an approval-gated tool call
```
curl -s -X POST http://localhost:8090/invoke \
  -H "Content-Type: application/json" \
  -d '{"message": "Can you block my card ending in 4321? I think it was stolen."}'
```
3.  Approve it once, expect a 200 response
```
curl -s -w "\nHTTP %{http_code}\n" -X POST http://localhost:8090/runs/<RUN_ID>/approvals/<APPROVAL_ID>/approve
```
4. Submit the exact same approval again, expect a 409 response
```
curl -s -w "\nHTTP %{http_code}\n" -X POST http://localhost:8090/runs/<RUN_ID>/approvals/<APPROVAL_ID>/approve
```

### Suggested Fix
Catch `ApprovalAlreadyProcessed` specifically and attempt recovery via `get_processed_result`. 
```
except ApprovalAlreadyProcessed as exc:
    try:
        recovered = await engine.get_processed_result(
            run_id, approval_id,
            caller_user_id=user_id, caller_session_id=caller_session_id,
        )
    except ApprovalError as recovery_exc:
        raise _map_approval_error(recovery_exc) from recovery_exc
    if recovered is None:
        raise _map_approval_error(exc) from exc
    result = recovered
```
I have a PR with a fix and tests ready, but can hold off if it is already being worked on.

Rsakhuja:
Fixes #109 
Catches `ApprovalAlreadyProcessed` exception and recovers the original result in order to avoid returning a 409 error when a duplicate request is made.


Asaf-prog:
Thanks for the fix — the overall approach looks correct, and recovering the already-processed result is consistent with the existing `agent_manager` behavior.

One issue I think we should address before merging:

If `engine.get_processed_result()` raises an unexpected exception that is **not** an `ApprovalError` (for example, a checkpointer/database/runtime failure), it will not be caught by the outer `except Exception` block, because it is raised from inside the `except ApprovalAlreadyProcessed` handler.

That means the recovery path currently bypasses the API's existing error sanitization behavior.

I suggest adding an `except Exception` around `get_processed_result()` that logs the failure and returns the standard sanitized `500 Internal server error`.

It would also be good to add a test where `get_processed_result()` raises something like `RuntimeError("private recovery failure")` and verify that:

* the response is `500`
* the response only contains `Internal server error`
* the internal exception message is not exposed to the client

Other than that, the change looks good to me.

Rsakhuja:
> Thanks for the fix — the overall approach looks correct, and recovering the already-processed result is consistent with the existing `agent_manager` behavior.
> 
> One issue I think we should address before merging:
> 
> If `engine.get_processed_result()` raises an unexpected exception that is **not** an `ApprovalError` (for example, a checkpointer/database/runtime failure), it will not be caught by the outer `except Exception` block, because it is raised from inside the `except ApprovalAlreadyProcessed` handler.
> 
> That means the recovery path currently bypasses the API's existing error sanitization behavior.
> 
> I suggest adding an `except Exception` around `get_processed_result()` that logs the failure and returns the standard sanitized `500 Internal server error`.
> 
> It would also be good to add a test where `get_processed_result()` raises something like `RuntimeError("private recovery failure")` and verify that:
> 
> * the response is `500`
> * the response only contains `Internal server error`
> * the internal exception message is not exposed to the client
> 
> Other than that, the change looks good to me.

Hi. That makes sense. I updated the PR to include the `except Exception` and the additional test.

AmitAvital1 (approved):

PR #111, opened 2026-08-20
Refactor/remove protocol call magic

Asaf-prog:
## Summary

This PR reorganizes several large and flat modules into focused, responsibility-oriented packages without intentionally changing runtime or API behavior.

### Runtime and application modules

- Split approval management, policies, providers, repositories, session approvals, and tool-execution persistence into focused modules.
- Split execution limits, execution state, hook loading, and tool-usage context responsibilities.
- Split engine approval capabilities into explicit interfaces.
- Split `ConversationService`, application errors, and prepared-turn data.
- Split authentication identity, token, key-source, and claim-mapping components.
- Extracted API schemas and generator result models from larger modules.

### LangGraph execution

- Preserved explicit node execution through:
  - `AgentNode.execute(...)`
  - `OrchestratorNode.execute(...)`
  - `builder.add_node(path, node.execute)`
  - `await entry.node.execute(...)`
- Removed callable-based node terminology and protocol indirection.
- Reorganized the LangGraph package into:
  - `nodes/`
  - `execution/`
  - `tools/`
  - `graph/`
- Extracted graph construction into `GraphBuilder`.
- Extracted MCP client construction and tool discovery into `MCPConnector`.
- Eliminated the generic `helpers.py` module by moving its functions into graph traversal, prompting, model context, execution context, and model-loop modules.
- Kept standalone checkpointing, approval-provider, and routing-filter modules at the package root.
- Verified the resulting import graph is acyclic.

### Conversation API routes

- Replaced the single large `agent_manager/api/routes.py` file with domain routers:
  - `routes/auth.py`
  - `routes/conversations.py`
  - `routes/approvals.py`
- Added a small `routes/router.py` aggregator.
- Moved shared HTTP error conversion into `api/errors.py`.
- Moved response projection and sanitization into `api/presenters.py`.
- Kept shared FastAPI dependencies in `api/deps.py`.
- Verified that all 11 API operations are registered exactly once.
- Compared the complete OpenAPI route contract before and after the refactor; it is unchanged.

### Skill adapter tooling

- Split skill synchronization and adapter generation into focused modules for:
  - adapter file handling
  - Claude targets
  - Codex targets
  - synchronization
  - shared target behavior
- Updated canonical and generated engineering guidance.
- Updated synchronization tests.

### Documentation

- Updated architecture, roadmap, execution-limit, MCP/tool, HITL, and tool-usage documentation to match the new module layout and terminology.

## Compatibility

This is intended to be behavior-preserving:

- No YAML contract changes.
- No public HTTP path, method, request, response, status, SSE, authorization, or OpenAPI changes.
- No changes to approval, checkpoint, resume, cancellation, tool execution, or MCP semantics.
- Internal imports using the previous module locations must use the new responsibility-oriented paths.

## Validation

- Ruff formatting: passed
- Ruff lint: passed
- Mypy: passed across 266 source files
- Pytest: 838 passed
- API/OpenAPI pre/post comparison: exact match
- Duplicate route registration audit: passed
- Import-cycle audits: passed
- Application and CLI startup imports: passed


AmitAvital1 (commented):
Read this as a refactor review: the mechanical part is solid — `ruff format`, `ruff check` and `mypy` are clean, the suite passes (the one `test_run_persistence` failure reproduces on `main`, unrelated), and I re-derived the import graph independently: acyclic. Every module path referenced in `docs/` still exists. `GraphBuilder` and `MCPConnector` coming out of `engine.py` is the right cut.

Two things I'd want resolved before merge, and both are about the "not breaking / staying simple" half of the goal rather than the file moves.

**1. `Protocol` → ABC is a behaviour change, not a move.** All 27 `typing.Protocol` uses are gone. `isinstance` against a `runtime_checkable` Protocol is structural; against an ABC it is nominal. Details inline on `conversation_service.py`, but the short version: one call site raises loudly, another silently takes the wrong branch. It also breaks any third-party `Engine` / `RunRepository` / `ApprovalProvider` that satisfies the shape without inheriting — which is what the Protocols were for. The PR additionally rewrites `.ai/skills/senior-python-engineering.md` to change the project standard from "Protocols/ABCs" to "ABCs", and `CLAUDE.md` says architecture decisions don't change without explicit approval. I'm not against the direction, but it wants its own ADR and its own PR, not a ride-along inside a reorganisation.

**2. The comments didn't survive the move.** 23 comment lines deleted, 5 added — and the deleted ones are the load-bearing ones: why the JWT algorithm allowlist exists, why `key_for` may read an unverified header, why `bool` is rejected as a user id, why `exhausted = True` sits inside `except`, what `begin_execution`'s bool means. The replacement docstrings restate the function name. Marked the ones I care about inline. Moving code is exactly when this context is most expensive to lose, because the next reader has no `git log` breadcrumb pointing at the original file.

**On the shape of the split.** 51 new files, ~30 of them under 40 lines — `token_error.py` is 5 lines, `token_claims.py` 6, `issued_token.py` 10. A 116-line `tokens.py` became 8 files and 192 lines; `anonymous_identity_source.py` now opens with 8 imports to support 30 lines of code. Meanwhile the two files that actually needed this went `engine.py` 1117 → 898 and `conversation_service.py` 631 → 598. The split took the easy value objects and left the hard ones. I'd merge `infrastructure/auth/` back into ~3 cohesive modules (`tokens.py`, `key_sources.py`, `identity.py`) — re-splitting later is more churn than getting it right once — and spend the difference on `engine.py`. Not blocking, but it's the difference between "simpler" and "more files".

**Nice to have, and I'd do these now since the lines are already in the diff:** the dead `_MODEL_FACTORY_OPTIONAL_KWARGS`, the `del state` parameters in `model_loop.py`, `MIN_SECRET_LENGTH` living in the contract module instead of beside its only user, `CLOCK_SKEW_LEEWAY_SECONDS` in a module whose docstring says "JWT claim names", and `context_models.py`'s docstring still claiming to own the selection logic that moved to `context_provider.py`.

**Nice to have, but please not in this PR:** the `"succeeded"` / `"streamable_http"` / `"mcp"` / `"local"` string literals where enums already exist, moving `graph/traversal.py` out of the LangGraph adapter into `core/` (it imports no LangGraph and `agentctl/diagnostics.py` now reaches into the adapter for it), a shared `ConversationError` base so `api/errors.py` can map by `isinstance`, and the SSE framing helper. All pre-existing, all real, and all of them would bury the refactor in unrelated diff.

src/agent_manager/application/conversation_service.py
```diff
@@ -12,26 +12,33 @@
...
-    ApprovalStreamingEngine,
-    Engine,
-    RunStatusEngine,
-)
-from agent_engine.engine.types import ChatMessage, RunResult
+from agent_engine.engine.approval_cancellation_engine import ApprovalCancellationEngine
+from agent_engine.engine.approval_engine import ApprovalEngine
+from agent_engine.engine.approval_streaming_engine import ApprovalStreamingEngine
+from agent_engine.engine.engine import Engine
+from agent_engine.engine.run_status_engine import RunStatusEngine
```
AmitAvital1, line 25: These five capability types moved from `runtime_checkable Protocol` to ABC, which flips `isinstance` from structural to nominal — that's a real behaviour change hiding in an import rename.

`_require_approval_engine` (L568) at least fails loudly. L528 is the one that worries me:

```python
elif isinstance(self._engine, RunStatusEngine):
```

An engine that implements `get_run_status` but doesn't inherit `RunStatusEngine` no longer matches, so we skip the branch silently and every message falls through to the `"completed" if message.run_id in completed else message.status` default. Wrong status, no exception, no log. That's the failure mode I'd least want in run lifecycle code.

If the intent is that engines must now inherit, let's say so in an ADR and make the mismatch loud here too.

src/agent_manager/infrastructure/auth/token_verifier.py
```diff
@@ -0,0 +1,57 @@
...
+        self._policy = policy or TokenPolicy()
+
+    def verify(self, token: str) -> Mapping[str, Any]:
+        policy = self._policy
+        try:
+            key = self._key_source.key_for(jwt.get_unverified_header(token))
+            claims: dict[str, Any] = jwt.decode(
+                token,
+                key,
+                algorithms=list(self._key_source.algorithms),
```
AmitAvital1, line 29: The comment that used to sit on this line is the one I'd fight hardest to keep:

```python
# The explicit allowlist is what rejects `alg: none` and
# algorithm confusion.
```

Without it this reads as plumbing, and the next person tidying up has no signal that it's the algorithm-confusion defence. Please carry it over verbatim.

src/agent_manager/infrastructure/auth/key_source.py
```diff
@@ -0,0 +1,18 @@
...
+
+from abc import ABC, abstractmethod
+from collections.abc import Mapping
+from typing import Any
+
+HMAC_SHA256 = "HS256"
+MIN_SECRET_LENGTH = 32
+
+
+class KeySource(ABC):
```
AmitAvital1, line 12: `KeySource` lost the docstring that made it safe to use — specifically that `key_for` receives the **unverified** header, so selecting on `kid` is only a hint and the signature check afterwards is what actually decides.

That sentence is the reason this class isn't a trust boundary. Without it, an implementer could reasonably conclude the header is trustworthy. Same for the module docstring explaining why key lookup is isolated at all (so an asymmetric issuer is a new class here and nothing else).

Minor while you're here: `MIN_SECRET_LENGTH` (L10) is only used by `static_secret_key_source.py` — it belongs beside its one caller, not in the contract module.

src/agent_manager/infrastructure/auth/claim_mapping.py
```diff
@@ -0,0 +1,56 @@
...
+            claims=claims,
+        )
+
+
+def _text(value: Any) -> str | None:
+    return value if isinstance(value, str) and value.strip() else None
+
+
+def _identifier(value: Any) -> str | None:
+    if isinstance(value, bool):
```
AmitAvital1, line 50: This guard now reads like a bug someone should fix. The deleted docstring is what stops that:

> JSON numbers are accepted because plenty of hosts key users on an integer primary key — Django and Rails out of the box. Booleans are not: `bool` is an `int` in Python, and a boolean id is a bug rather than an identity.

Two lines of comment, and the `isinstance(value, bool)` check stops looking arbitrary.

src/agent_manager/infrastructure/auth/identity_resolver.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def resolve(self, token: str) -> Principal:
+        if _is_anonymous_pass(token):
+            return self._anonymous.resolve(token)
+        if self._host is None:
+            raise TokenError("no host identity is configured; only visitor passes are accepted")
+        return self._host.resolve(token)
+
+    def names_a_host_user(self, token: str) -> bool:
+        try:
```
AmitAvital1, line 36: A method named `names_a_host_user` that reads an unverified JWT header really wants the note that was here: it only picks between two tokens the caller already supplied, it grants nothing on its own, and whichever token wins still goes through `resolve` where the signature check decides. A token too malformed to read is deliberately not treated as a claim.

Right now the `except TokenError: return False` looks like swallowing an error rather than a decision.

src/agent_manager/api/routes/conversations.py
```diff
@@ -0,0 +1,165 @@
...
+            ).model_dump(exclude_none=True)
+            yield f"event: turn_started\ndata: {json.dumps(started)}\n\n"
+            async for event in stream:
+                suspended = suspended or event.type == "pending_approval"
+                payload = to_stream_event(event).model_dump(exclude_none=True)
+                yield f"event: {event.type}\ndata: {json.dumps(payload)}\n\n"
+            exhausted = True
+        except Exception:
+            await service.fail_turn(turn)
+            exhausted = True
```
AmitAvital1, line 152: This now reads as a redundant assignment — `exhausted` is already set on the success path, so why here. The original said it:

```python
# The run is already terminal; `finally` must not try to cancel it.
```

Without that, someone deletes this line as dead code and we start double-cancelling a turn that already failed.

src/agent_manager/api/routes/conversations.py
```diff
@@ -0,0 +1,165 @@
...
+            logger.exception("conversation stream failed")
+            payload = {"type": "error", "error": INTERNAL_ERROR_MESSAGE}
+            yield f"event: error\ndata: {json.dumps(payload)}\n\n"
+        finally:
```
AmitAvital1, line 156: Same story — the nested `try`/`finally` shape is deliberate and the comment explaining it went away:

```python
# StreamingResponse cancellation closes the application-owned
# generator, which in turn cancels and awaits the graph producer.
```

That's the whole reason `stream.aclose()` and `cancel_turn` are ordered the way they are.

src/agent_manager/api/routes/approvals.py
```diff
@@ -0,0 +1,158 @@
...
+            result = await service.decide_approval(
+                conversation_id,
+                run_id,
+                approval_id,
+                body.decision,
+                caller,
+            )
+    except HTTPException:
+        raise
+    except ApprovalError as exc:
```
AmitAvital1: This `HTTPException` → `ApprovalError` → `Exception`/500 ladder is now byte-identical in three places: here, `cancel_pending_approval` (L143), and a shortened variant in `conversations.py` L109. Splitting one file into three turned one block into three copies that will drift.

The PR already established the pattern for this — `as_http_error` in `api/errors.py`. A sibling there does it:

```python
@contextmanager
def as_approval_http_error(action: str, *, run_id: str, approval_id: str) -> Iterator[None]:
    ...
```

Collapses roughly 60 lines and keeps the three endpoints agreeing on what a rejected approval looks like over the wire.

src/agent_engine/approvals/tool_execution_manager.py
```diff
@@ -0,0 +1,48 @@
...
+        if self._executions is None:
+            return None
+        record = await self._executions.get(execution_id)
+        if record is not None and record.status == "succeeded":
+            return record
+        return None
+
+    async def begin_execution(
+        self, execution_id: str, *, tool_call_id: str, run_id: str, tool_name: str
+    ) -> bool:
```
AmitAvital1, line 43: `-> bool` on a method called `begin_execution` is genuinely ambiguous — I'd read `False` as "failed to begin". The docstring that disambiguated it didn't make the move:

> Claim the idempotency key. Returns True if this attempt should run, False if another attempt already owns it (duplicate).

Given this is the deduplication path that stops a second real side effect after resume, that sentence is worth more than the one-liner on the class.

The class docstring also dropped "Holds no per-run mutable state; safe to share across concurrent runs" — that's a concurrency invariant callers rely on when they share one instance across runs.

src/agent_engine/approvals/tool_execution_manager.py
```diff
@@ -0,0 +1,48 @@
...
+from __future__ import annotations
+
+import hashlib
+
+from agent_engine.approvals.models import ToolExecutionRecord
+from agent_engine.approvals.tool_execution_repository import ToolExecutionRepository
+
+
+def execution_id_for(tool_call_id: str, *, salt: str = "") -> str:
+    """Return a deterministic idempotency key for one tool call."""
```
AmitAvital1: The docstring here lost the only part that mattered — *why* the key is derived from `tool_call_id`:

> Derived from the stable `tool_call_id` so that a retry or a graph re-entry which reaches the same tool call computes the same key and can be deduplicated. Optional `salt` (e.g. an approval id) scopes it further.

"Return a deterministic idempotency key" just restates the signature; the replaced text explained the property the whole HITL resume path depends on, and what `salt` is for.

src/agent_engine/engine/langgraph/graph/graph_builder.py
```diff
@@ -0,0 +1,247 @@
...
+from agent_engine.engine.langgraph.tools.tool_invoker import ToolInvoker
+from agent_engine.loaders.resolver_loader import ResolverLoader
+from agent_engine.loaders.tool_loader import ToolLoader
+from agent_engine.runtime.hooks import HookManager
+from agent_engine.runtime.state import GraphState
+from agent_engine.tool_usage.context_provider import ToolUsageContextProvider
+from agent_engine.tool_usage.tracker import ToolUsageTracker
+
+ModelFactory = Callable[..., BaseChatModel]
+_MODEL_FACTORY_OPTIONAL_KWARGS = ("region", "max_tokens", "top_p")
```
AmitAvital1, line 35: `_MODEL_FACTORY_OPTIONAL_KWARGS` isn't referenced anywhere — and the same three keys are spelled out again as literals in the dict at L47-51. It was already dead in `engine.py`; carrying it into a new file is the moment to either use it:

```python
optional = {key: getattr(model, key) for key in _MODEL_FACTORY_OPTIONAL_KWARGS}
```

or drop it.

src/agent_engine/engine/langgraph/execution/model_loop.py
```diff
@@ -0,0 +1,121 @@
...
+
+async def _refresh(
+    context: ModelContext,
+    refresh_execution_context: ExecutionContextRefresher | None,
+) -> None:
+    if refresh_execution_context is not None:
+        context.set_execution_context(await refresh_execution_context())
+
+
+def emit_route(state: GraphState, route: tuple[str, ...]) -> None:
```
AmitAvital1: `del state` as the first statement is the function telling us the parameter is dead. Since this is a fresh module boundary, better to just drop it from the signature — and the same applies to `state` in `invoke_model` (L26), which never touches it, and therefore to the `state` argument `run_tool_loop` only forwards.

Three signatures get shorter and nothing calls them from outside this package.

src/agent_engine/engine/langgraph/filters.py
```diff
@@ -36,7 +21,8 @@ class RouteFilter(ABC):
     """
 
     @abstractmethod
-    def filter(self, ctx: dict[str, Any], candidates: list[T]) -> list[T]: ...
+    def filter(self, ctx: dict[str, Any], candidates: list[ChildEntry]) -> list[ChildEntry]:
```
AmitAvital1: `RouteFilter` is documented right above as the extension point for "access control, feature flags, rate limiting, etc. without touching the engine core" — but dropping `Filterable` and the `TypeVar` nails it to `ChildEntry`, a concrete type from `nodes/`. The docstring that was deleted said exactly why the Protocol existed:

> Decouples filters from the spec layer (`GraphNode`) and the runtime layer (node callables): a filter only ever inspects a candidate's id and whether it is protected.

Now an extension point in the adapter's public surface depends on an internal runtime type, and the `TYPE_CHECKING` import is a hint that the layering isn't quite right. If Protocols are going away project-wide, a two-method ABC `Filterable` keeps the same decoupling.

src/agent_engine/loaders/resolver_loader.py
```diff
@@ -30,15 +29,21 @@ def __init__(self, base_dir: Path) -> None:
         self._instances: dict[str, Any] = {}
         self._shared_loaded = False
 
-    def load(self, agent_id: str, resolver_id: str) -> Callable[[dict[str, Any]], Any]:
+    def resolve(
```
AmitAvital1: `load(agent_id, resolver_id) -> Callable` became `resolve(agent_id, resolver_id, context) -> Any`. I like the change — Tell-Don't-Ask, and the caller no longer holds a bound method it might stash — but it's a public loader API rename with a new parameter, and the PR description says no public contract changed. Worth calling out explicitly there so nobody's plugin breaks on an upgrade without warning.

src/agent_engine/api/schemas.py
```diff
@@ -0,0 +1,57 @@
...
+    arguments: dict[str, Any] = {}
+
+
+class InvokeResponse(BaseModel):
+    system_name: str
+    answer: str
+    visited: list[str]
+    used_tools: list[ToolRecord]
+    run_id: str
+    status: str = "completed"
```
AmitAvital1: New file, so this is the cheap moment: `status: str = "completed"` where `RunStatus.COMPLETED` already exists in `approvals/models.py`. `RunStatus` is a `StrEnum`, so `status: RunStatus = RunStatus.COMPLETED` serialises identically and the API stops being able to emit a status the engine can't produce.

Same for `RunStatusResponse.status` below.

.ai/skills/senior-python-engineering.md
```diff
@@ -41,8 +41,8 @@ future agents can safely modify without fear.
   in adapters at the edges; core logic stays pure and testable.
 - **Dependency injection over hidden globals.** Pass collaborators in; no
   module-level mutable singletons, and never mutable global request state.
-- **Adapters via Protocols/interfaces.** External integrations (LLM, MCP, DB,
-  plugins) sit behind `typing.Protocol`/ABCs so they can be faked in tests.
+- **Adapters via explicit interfaces.** External integrations (LLM, MCP, DB,
+  plugins) sit behind ABCs so implementations are navigable and can be faked in tests.
```
AmitAvital1: This is the one I'd fix before merge, because it's the file that tells the next contributor (and the next agent) what to write.

`7d49a50` correctly put the engine capabilities back on `runtime_checkable Protocol`, and `c3a071f` restored `Filterable` — so the PR now ships five Protocols while this line, L62 and L85 all say external integrations sit behind ABCs. The standard and the code disagree inside the same commit range.

Reverting these three lines to the original "Protocols/ABCs" wording (plus `make generate-ai`, since `.claude/skills/` and `.codex/skills/` carry the same text) is the whole fix. If the intent is genuinely "ABCs for ports, Protocols for optional capability detection" — which is what the code now does, and is a sound rule — that's worth stating explicitly here rather than leaving the two in contradiction.

src/agent_manager/infrastructure/auth/claim_mapping.py
```diff
@@ -0,0 +1,57 @@
...
+            claims=claims,
+        )
+
+
+def _text(value: Any) -> str | None:
+    return value if isinstance(value, str) and value.strip() else None
+
+
+def _identifier(value: Any) -> str | None:
+    """Return a usable textual identity claim; booleans are never identities."""
```
AmitAvital1: Half of it landed — "booleans are never identities" is the conclusion, but the two facts a reader needs are still missing: *why* `int` is accepted at all, and *why* `bool` needs its own branch.

The original had both:

> JSON numbers are accepted because plenty of hosts key users on an integer primary key — Django and Rails out of the box. Booleans are not: `bool` is an `int` in Python, and a boolean id is a bug rather than an identity.

The "`bool` is an `int` in Python" half is the actual trap — without it the `isinstance(value, bool)` branch looks defensive rather than necessary, and someone will eventually merge it into the `int` check.

src/agent_engine/approvals/tool_execution_manager.py
```diff
@@ -0,0 +1,49 @@
...
```
AmitAvital1: `begin_execution` got its docstring (thanks) but this one didn't move with it, and it's the more load-bearing of the two — it explains the property the whole HITL resume path rests on:

> Derived from the stable `tool_call_id` so that a retry or a graph re-entry which reaches the same tool call computes the same key and can be deduplicated. Optional `salt` (e.g. an approval id) scopes it further.

As it stands, "Return a deterministic idempotency key for one tool call" restates the signature, and `salt` is undocumented.

src/agent_engine/approvals/tool_execution_manager.py
```diff
@@ -0,0 +1,49 @@
...
+    digest = hashlib.sha256(f"{tool_call_id}:{salt}".encode()).hexdigest()
+    return f"exec_{digest[:24]}"
+
+
+class ToolExecutionManager:
+    """Coordinates idempotent tool execution through an injected repository."""
```
AmitAvital1: Also still missing here: *"Holds no per-run mutable state; safe to share across concurrent runs."*

That's not decoration — it's the invariant that lets `GraphBuilder` hand one instance to every `ToolInvoker` in the graph. Without it, someone looking at concurrent runs has to re-derive it from the fact that the only attribute is the injected repository.

src/agent_manager/api/routes/conversations.py
```diff
@@ -0,0 +1,164 @@
...
+                yield f"event: {event.type}\ndata: {json.dumps(payload)}\n\n"
+            exhausted = True
+        except Exception:
+            await service.fail_turn(turn)
+            # The run is already terminal; `finally` must not try to cancel it.
+            exhausted = True
+            logger.exception("conversation stream failed")
+            payload = {"type": "error", "error": INTERNAL_ERROR_MESSAGE}
+            yield f"event: error\ndata: {json.dumps(payload)}\n\n"
+        finally:
```
AmitAvital1, line 156: The `except` comment came back but this one didn't. The nested `try`/`finally` here is deliberate and non-obvious:

```python
# StreamingResponse cancellation closes the application-owned
# generator, which in turn cancels and awaits the graph producer.
```

It's the reason `stream.aclose()` has to complete before the `cancel_turn` decision, rather than the two sitting side by side.

AmitAvital1 (commented):
Went through all ten commits against the original comments — this is a good round. Re-ran the gate on the new head: `ruff format`, `ruff check`, `mypy` (270 files) and the suite all pass (the `test_run_persistence` failure still reproduces on `main`, unrelated), and the import graph is still acyclic.

Eleven of fifteen are fully resolved, and the two that actually blocked are among them:

- `7d49a50` puts the four optional capabilities back on `runtime_checkable` Protocol, so `isinstance` capability detection is structural again and the silent wrong-branch at `_with_run_statuses` is gone.
- The security context came back on `token_verifier`, `key_source`, `identity_resolver` and the stream `except` — and came back *shorter* than the originals, which is the right call. The invariant survived, the verbosity didn't.
- `8c1e252` is better than what I suggested: splitting it into `as_approval_http_error` + `as_internal_http_error` makes the two concerns composable instead of one do-everything wrapper, and the nesting order keeps `ConversationNotFound` mapping to 404 before the approval handler sees it. `approvals.py` lost 20 lines and the three endpoints can't drift apart now.
- `143adcd` typing `RunStatus` end to end **plus** a test pinning the OpenAPI shape to an unrestricted string is exactly the careful version of that change.
- `fda6dd2` adding resolver caching coverage wasn't asked for and is welcome.

Five things left. None of them are correctness, all of them are cheap, and I've marked four of them inline. Happy to approve as soon as they're in — or to approve now and take them as a follow-up if you'd rather not hold the branch.

The fifth is the PR description: it still says no public contract changed, but `ResolverLoader.load(agent_id, resolver_id) -> Callable` is now `resolve(agent_id, resolver_id, context) -> Any`. Good change, just worth a line in the body so nobody's plugin discovers it on upgrade.

src/agent_engine/engine/langgraph/engine.py
```diff
@@ -218,28 +191,6 @@ def _pending_approval_event(system_name: str, result: RunResult) -> RunStreamEve
...
-    except (TypeError, ValueError):
-        return present
-    parameters = signature.parameters.values()
-    if any(param.kind == inspect.Parameter.VAR_KEYWORD for param in parameters):
-        return present
-    accepted = set(signature.parameters)
-    return {key: value for key, value in present.items() if key in accepted}
-
-
 class LangGraphEngine(Engine):
```
AmitAvital1, line 194: Not a regression — `main` had the same exposure — but the PR is the moment, and this round made it slightly sharper.

Now that the capabilities are Protocols again, `LangGraphEngine` doesn't inherit them, and nothing pins that it still satisfies them: `isinstance` against a `runtime_checkable` Protocol only checks that the attribute names exist, not the signatures, and there's no explicit conformance assertion for mypy to check either. `143adcd` just changed `get_run_status` from `-> str` to `-> RunStatus` — exactly the kind of drift that would pass silently here and degrade the manager to a wrong-status fallback at runtime.

One line in the engine tests closes it for good:

```python
assert isinstance(engine, ApprovalEngine | ApprovalStreamingEngine | ApprovalCancellationEngine | RunStatusEngine)
```

or four separate asserts if you prefer the failure message to name which one drifted.

AmitAvital1 (approved):
All five code/doc items from the last round are in, and two of them came back better than what I asked for.

`tests/engine/test_engine_capabilities.py` is the one I want to call out. I suggested a single `isinstance` assert; the annotated binding + assert pair is the more complete answer, because it covers the two different failure modes separately — `isinstance` for what `ConversationService` actually does at runtime (attribute names only), and the annotation for the signature drift `isinstance` is blind to. I checked that it bites rather than just passing: reverting `get_run_status` to `-> str` makes mypy fail on the annotated binding with the exact `Coroutine[Any, Any, RunStatus]` vs `Coroutine[Any, Any, str]` mismatch. Four separate tests so the failure names which capability drifted is the right touch too.

The skill file is also better than the revert I proposed. Stating the actual rule — ABCs for ports, `runtime_checkable` Protocols for optional capabilities detected with `isinstance`, plain Protocols for the minimal shape a function needs — turns what was a contradiction into guidance someone can follow, and regenerating `.claude/` and `.codex/` closes it properly.

Gate on `bf9d25b`: `ruff format` and `ruff check` clean, `mypy` clean across 271 files, suite green. The one `test_run_persistence` failure reproduces identically on `origin/main` at `aae8238`, so it isn't this branch. Import graph still acyclic.

Approving. One loose end that isn't worth another round: the PR description still says no public contract changed, and `ResolverLoader.load(agent_id, resolver_id) -> Callable` → `resolve(agent_id, resolver_id, context) -> Any` is still unmentioned. Worth a line in the body before merge, especially if this lands as a squash and the description becomes the commit message.

On the module fragmentation I raised earlier — I'm not holding the branch for it. It's a judgement call about shape, the split is coherent as it stands, and `infrastructure/auth/` can be consolidated later without touching behaviour if it starts to grate.

PR #112, opened 2026-08-20
feat(auth): carry the caller's own credential through to plugin code

AmitAvital1:
## Why

A tool had no way to reach the identity its request arrived with. Any integration acting on a host product therefore had to hold one shared privileged key — which silently grants every caller whatever that key can do. In the Open WebUI example this meant an ordinary user could perform admin actions through chat.

`AuthContext.inbound_access_token` was built for exactly this and documented as *"the host app passes it in"*, but no HTTP path ever reached it, so it was always `None`.

## What

Fill it on each route that has a caller:

- **Manager** attaches the token once its signature verifies, so nothing downstream sees a credential this process did not just check. It rides on the prepared turn, because `stream_turn` executes with nothing else in hand.
- **Engine `/invoke` and `/stream`** forward `Authorization` verbatim, without verifying it. A forged credential fails at the system that owns it, so passing one along grants nothing that direct access would not. Facts the caller merely *asserts* (a user id, a role) get no such treatment and are not accepted — which is why no context object was added to the request body.
- **Both approval resume paths** take the approver's credential as of the decision, not one captured when the run started and possibly expired while it waited for a human.

Visitor passes are never carried: they prove nothing to a host product.

## Boundary

The engine is a conduit; the manager is an authority. Both produce the same `RunContext`. Nothing about how a credential is obtained, refreshed, or exchanged lives here — that is plugin code, reached through the existing `on_run_start` hook, which may already replace the context.

## Safety

- `inbound_access_token` never enters graph state, so it cannot reach a prompt, the model, or a response — pinned by a test.
- `repr=False` keeps it out of logs and tracebacks.
- `Principal` is never serialized wholesale (only `ToolUsageRecord` goes through `asdict`).

## Docs

Two published pages were missing what a user needs to actually do this:

- `mcp-and-tools.mdx` — how a tool acts as the caller, placed right after "Implementing a tool" where someone writing one arrives at the question.
- `runtime-hooks.mdx` — `plugins.import_roots`, without which managed hooks fail to import. The page walked you through registration and stopped one step short; the warning now carries the exact error string.

## Tests

Six added. Token survives resolution; a visitor pass never carries one; it stays out of `repr`; `Bearer` parsing including case and the empty cases; the approver's credential reaches resume; and the graph-state invariant.

`make check` is clean. `tests/cli/test_run_persistence.py::test_agentctl_run_without_session_prints_reusable_generated_session` fails identically on a clean `main` (a click version issue in that test's harness) and is unrelated.

### DIFF
--- a/docs/MCP_AND_TOOLS.md
+++ b/docs/MCP_AND_TOOLS.md
@@ -410,8 +410,9 @@ For the MVP:
 - hide local-vs-MCP origin by presenting both as LangChain tools, while tracking
   the origin in the tool-usage records (✅ implemented);
 - bind discovered MCP tools into LangGraph/LangChain tool-calling (✅ implemented);
-- pass trusted request context through `ctx` into tool calls (⏳ planned —
-  resolvers receive `ctx`, tools do not yet);
+- reach the run's context from a tool via `current_run_context` (✅ implemented —
+  a context variable rather than a `ctx` parameter, so tool signatures stay
+  exactly what the model sees);
 - redact secrets from traces (⏳ planned, task 0011);
 - keep prompt wording out of the enforcement path.
 
--- a/docs/SIDECAR_CONTEXT_AUTH.md
+++ b/docs/SIDECAR_CONTEXT_AUTH.md
@@ -27,7 +27,41 @@ Content-Type: application/json
 
 For each request, the runtime builds `ctx` from headers and request data. Plugin
 methods receive `ctx`; client code decides how to interpret tokens, user ids,
-tenant ids, roles, and permissions.
+tenant ids, roles, and permissions. Tools are not passed `ctx` as an argument —
+that would put it in the schema the model sees — and read it from
+`agent_engine.runtime.hooks.current_run_context` instead.
+
+---
+
+## Credentials Are Accepted, Facts Are Derived
+
+`Authorization` reaches `ctx.auth_context.inbound_access_token` and is forwarded
+to plugin code without being verified here. That is deliberate, and the split it
+rests on is the one rule to keep in mind:
+
+- **A credential is self-securing.** Forge it and the system that owns it answers
+  401. Nothing can reach through this engine that it could not already reach
+  directly, so taking one from the caller costs nothing.
+- **A fact is not.** `{"role": "admin"}` works the moment anything trusts it.
+
+So a credential may be accepted from the caller; a user id, a role, or a tenant
+must be *derived* — a resolver calls the owning system with the credential and
+uses what comes back. Nothing else about the caller is accepted over the wire,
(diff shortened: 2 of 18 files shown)


Asaf-prog:
Thanks — the overall direction looks good, but I found one blocking issue before I can approve.

The token is correctly forwarded through `/invoke` and `/stream`, and the engine resume APIs now support `access_token`. However, the HTTP approval endpoints (`/decision`, `/approve`, `/reject`) still do not read the `Authorization` header or pass the current approver token into `engine.resume()`.

This means a flow like:

`/invoke` with a user token → pending approval → `/approve` → resumed tool execution

will resume without `ctx.auth_context.inbound_access_token`, so the tool cannot act as the user after approval.

I think we should:

* Accept `Authorization` on the approval endpoints.
* Pass the current approver's bearer token to `engine.resume(..., access_token=...)`.
* Add a test proving that a run started with `TOKEN_A` and approved with `TOKEN_B` resumes using `TOKEN_B`.

One small documentation issue as well: `AuthContext` still says that the platform never populates `inbound_access_token`, which is no longer true after this change.

Once the approval path is covered, this looks much closer to approval.

AmitAvital1:
Good catch — fixed in 3593608.

`/decision`, `/approve`, and `/reject` now read `Authorization` the same way `/invoke`/`/stream` already do, and forward it to `engine.resume(access_token=...)`. Added `test_approval_resumes_with_the_approvers_token_not_the_original_caller`: a run started with `TOKEN_A` and approved with `TOKEN_B` resumes with `TOKEN_B`.

Also fixed the `AuthContext` docstring — it still said the platform never populates `inbound_access_token`, which stopped being true once `/invoke`/`/stream` started filling it from the request.

Asaf-prog:
Thanks, the approval flow issue is fixed now and the new `TOKEN_A → approve with TOKEN_B → tool sees TOKEN_B` test covers it well.

I found one small security issue before approving:

`AuthContext.inbound_access_token` is currently a regular dataclass field, so the token can appear in `repr(ctx.auth_context)` or even `repr(ctx)`.

Since this is a credential, I think it should get the same protection already used for `Principal.access_token`:

```python
inbound_access_token: str | None = field(default=None, repr=False)
```

It would also be good to add a test asserting that the token is not present in either `repr(ctx.auth_context)` or `repr(ctx)`.

After that, I’m good with approving this.

AmitAvital1:
Fixed in bec5c30 — `inbound_access_token` now has `repr=False`, matching `Principal.access_token`. Added `test_the_token_is_not_in_auth_context_repr` and `test_the_token_is_not_in_run_context_repr`.

Asaf-prog (approved):

PR #124, opened 2026-08-22
feat(agent_manager): add pagination and lazy loading for chat history (#116)

### LINKED ISSUE #116: Add Pagination and Lazy Loading to Chat History
### Problem

The Chat History UI already exists, but currently it loads **all conversations at once**.

This does not scale well as the number of conversations grows and causes unnecessary:

* Network traffic.
* Backend work.
* UI rendering.
* Memory usage.
* Initial loading time.

We should change the existing Chat History implementation so conversations are loaded incrementally instead of fetching the entire history.

## Proposed Change

Implement pagination in the Chat History API and lazy loading in the UI.

When the Chat History is opened, the UI should fetch only the first page of conversations.

For example:

```http
GET /conversations?limit=20
```

As the user scrolls toward the bottom of the list, the UI should dynamically request the next page:

```http
GET /conversations?limit=20&cursor=...
```

The new conversations should be appended to the existing list without reloading the entire history.

### Expected Flow

```text
Open Chat History
        ↓
Load first page only
        ↓
Render conversations
        ↓
User scrolls down
        ↓
Near bottom of loaded list
        ↓
Fetch next page
        ↓
Append conversations
        ↓
Continue until no more results
```

The UI should **never fetch all conversations upfront**.

## Ordering

The current ordering should remain based on the latest activity in the conversation.

The conversation where the most recent message was written should appear first:

```text
ORDER BY last_message_at DESC
```

Pagination must preserve this ordering.

Prefer a deterministic secondary ordering field as well, for example:

```text
ORDER BY last_message_at DESC, id DESC
```

This prevents unstable pagination when multiple conversations have the same `last_message_at` value.

## Pagination

Prefer cursor-based pagination rather than offset-based pagination, since the history can change while the user is scrolling.

Example response:

```json
{
  "items": [
    {
      "id": "conversation-123",
      "title": "Invoice estimation",
      "lastMessageAt": "2026-08-20T20:15:00Z"
    }
  ],
  "nextCursor": "..."
}
```

When `nextCursor` is `null`, the UI knows there are no additional conversations to load.

## UI Behavior

The loading should feel seamless.

When loading the next page:

* Existing conversations must remain visible.
* Show a small loading indicator at the bottom of the list.
* Do not block the entire Chat History UI.
* Do not trigger multiple concurrent requests for the same next page.
* Append new conversations to the existing list.
* Avoid duplicate conversations.
* Stop requesting pages once the end of the history is reached.

## Important

This issue is specifically an **improvement to the existing Chat History implementation**.

We are not adding a new Chat History feature.

The goal is to replace:

```text
Open Chat History
        ↓
Fetch ALL conversations
        ↓
Render everything
```

with:

```text
Open Chat History
        ↓
Fetch first page
        ↓
Render
        ↓
Scroll
        ↓
Fetch next page dynamically
```

## Acceptance Criteria

* Existing Chat History API is changed to support pagination.
* Opening Chat History does not fetch all conversations.
* Only the first page is fetched initially.
* Additional pages are fetched dynamically when the user scrolls.
* Newly fetched conversations are appended to the existing list.
* Conversations remain ordered by latest message/activity descending.
* Pagination ordering is deterministic.
* The UI does not issue duplicate pagination requests.
* The same conversation is not rendered twice.
* A bottom loading state is shown while fetching another page.
* Pagination stops when there are no more conversations.
* Existing Chat History behavior for selecting/opening a conversation continues to work.
* Tests cover backend pagination, ordering, and UI lazy-loading behavior.

rishu685:
### Summary of Changes

Closes #116

- **Domain Models & Repository Port**: Added `PaginatedSessions` dataclass and updated `Repository.list_sessions(user_id, limit=50, cursor=None)`.
- **Persistence Layer**:
  - `sql_repository.py`: Implemented opaque URL-safe base64 cursors, `ORDER BY last_message_at DESC NULLS LAST, session_id DESC`, and `limit + 1` row fetching for `next_cursor` determination.
  - `memory_repository.py`: Implemented matching cursor filtering and sorting logic.
- **Application & API Layer**:
  - Updated `ConversationService.list_conversations` to return `PaginatedSessions`.
  - Updated `GET /conversations` endpoint with query parameters `limit` (default 20, 1-100) and `cursor`, returning `PaginatedConversationsResponse`.
- **React Widget UI**:
  - `AgentChatClient.ts` & `useConversation.ts`: Added `limit` and `cursor` pass-throughs.
  - `AgentChatApp.tsx`: Implemented infinite scroll in `ThreadDrawer` with scroll proximity triggers, deduplication, and loading indicator.
- **Tests**: Added `tests/agent_manager/test_pagination.py` covering cursor encoding, multi-page database navigation, ordering, and API endpoints. Updated existing test suite for paginated response envelopes. All 679 tests passing.


src/agent_manager/infrastructure/persistence/memory_repository.py
```diff
@@ -14,14 +14,30 @@
     ConversationSession,
     ConversationSnapshot,
     Message,
+    PaginatedSessions,
     Repository,
     Role,
     User,
 )
+from agent_manager.infrastructure.persistence.sql_repository import (
```
AmitAvital1: `MemoryRepository` is the default, dependency-free adapter, and this makes it import the SQLAlchemy adapter module to get a base64/JSON codec, pulling in `sqlalchemy`, `sqlmodel` and the table registry with it. Two adapters behind the same port shouldn't know about each other; that independence is what `test_repository_contract.py` exists to prove.

It doesn't break today only because `sqlalchemy` is an unconditional dependency.

The concrete cost is the timezone bug on the `sql_repository.py:221` thread: the cursor codec is a contract shared between adapters but it lives inside one of them, so nobody owns normalizing its format, and the two ended up emitting incompatible encodings of the same value. Moving it to `persistence/pagination.py` gives the contract an owner.

src/agent_manager/api/routes/conversations.py
```diff
@@ -45,17 +46,23 @@ async def create_conversation(
...
-    sessions = await service.list_conversations(caller)
-    return [
+@router.get("/conversations", response_model=PaginatedConversationsResponse)
+async def list_conversations(
+    service: Service,
+    caller: Caller,
+    limit: int = Query(default=20, ge=1, le=100),
+    cursor: str | None = Query(default=None),
+) -> PaginatedConversationsResponse:
+    paginated = await service.list_conversations(caller, limit=limit, cursor=cursor)
```
AmitAvital1: This is the only route in the file that isn't wrapped in `as_http_error()`. `decode_cursor` raises a bare `ValueError`, nothing maps it, and it escapes to Starlette's default handler:

```
cursor=garbage!!!                     -> 500
cursor=                               -> 500
cursor=b64('{"t": 12345, "id": "x"}') -> 500
cursor=b64('[1,2,3]')                 -> 500
```

Note the second case. An empty `?cursor=` is `""`, not `None`, so it passes the `is not None` check in both adapters and reaches the decoder. The widget never sends it because of the `if (cursor)` guard in `AgentChatClient.ts:73`, but this is a public endpoint and an empty query parameter is the ordinary way a client produces one.

A truncated or stale cursor is invalid client input, not a server fault. Raise a typed error from the codec, register it in `_HTTP_ERRORS`, and wrap the route body like its five neighbours. The response then carries the `{error_type, message}` shape the widget already parses in `AgentChatClient.ts:28`.

`test_pagination.py:46` tests the decoder in isolation, so it passes while the endpoint 500s. A test asserting the status code would have caught this.

AmitAvital1 (commented):
Review notes on the pagination change. Full summary is in the review that requests changes; details are inline.

src/agent_manager/infrastructure/persistence/sql_repository.py
```diff
@@ -36,6 +39,27 @@
...
+
+def decode_cursor(cursor_str: str) -> tuple[datetime | None, str]:
+    try:
+        raw_bytes = base64.urlsafe_b64decode(cursor_str.encode("ascii"))
+        data = json.loads(raw_bytes.decode("utf-8"))
+        t_str = data.get("t")
+        last_message_at = datetime.fromisoformat(t_str) if t_str is not None else None
+        session_id = str(data["id"])
+        return last_message_at, session_id
+    except Exception as exc:
```
AmitAvital1: `except Exception` also swallows genuine bugs. An `AttributeError` or `TypeError` introduced by a later refactor inside the `try` would be reported to the caller as an invalid cursor, and the real fault would be invisible.

Narrow it to what decoding can raise: `(binascii.Error, UnicodeDecodeError, json.JSONDecodeError, ValueError, KeyError, TypeError)`.

Line 60 also interpolates the client-supplied cursor into the exception message, which then reaches the logs verbatim. Drop the value.

src/agent_manager/infrastructure/persistence/sql_repository.py
```diff
@@ -151,16 +175,58 @@ async def get_session(self, session_id: str) -> ConversationSession | None:
...
+                )
+
         stmt = (
             select(ConversationSessionRow)
-            .where(ConversationSessionRow.user_id == user_id)
-            .order_by(col(ConversationSessionRow.last_message_at).desc())
-            .limit(limit)
+            .where(*conditions)
+            .order_by(
+                col(ConversationSessionRow.last_message_at).desc().nullslast(),
```
AmitAvital1: `nullslast()` changes the ordering contract. The previous key was `last_message_at or created_at`; dropping the `created_at` fallback leaves every session with no messages tied on the first two key components, ordered only by `session_id`, which is a `uuid4().hex`. On six sessions created in a known order with no messages:

```
matches newest-first?      : False
is it just session_id desc?: True
```

So empty conversations come back in arbitrary order, which contradicts the port's docstring one file over ("most-recently-active first") and regresses the default backend. `create_session` never sets `last_message_at`, so this is the normal state of a conversation between creation and its first message.

`memory_repository.py:133-140` carries the same change and has to move with whatever is decided here.

The fix I'd suggest is keying both adapters on `COALESCE(last_message_at, created_at)`. It restores the old ordering, keeps the backends aligned, makes the cursor's `t` non-nullable, and removes the NULL branch from both predicates.

Worth noting the test at `test_pagination.py:122` asserts `set(fetched_ids) == {c1, c2, c3}` over exactly three empty conversations, so it discards the order it looks like it's checking. `assert fetched_ids == [c3, c2, c1]` would pin the contract instead.

src/agent_manager/infrastructure/persistence/sql_repository.py
```diff
@@ -151,16 +175,58 @@ async def get_session(self, session_id: str) -> ConversationSession | None:
...
+                col(ConversationSessionRow.session_id).desc(),
+            )
+            .limit(limit + 1)
```
AmitAvital1, line 191: This query filters on `user_id` and orders by `(last_message_at DESC, session_id DESC)`, and no index serves that combination. `conversation_sessions` has separate single-column indexes on `user_id` and `last_message_at`. `EXPLAIN QUERY PLAN`:

```
SEARCH conversation_sessions USING INDEX ix_conversation_sessions_user_id (user_id=?)
USE TEMP B-TREE FOR ORDER BY
```

Every page sorts the user's entire session set from scratch. For the heavy user this feature exists for, 5,000 conversations over 250 page requests means 250 full sorts where the unpaginated query did one, so the change currently makes the total work worse. The point of keyset pagination is the index-only seek that makes page N cost the same as page 1.

`idx_conversation_messages_session_created` in `tables.py:57` shows the pattern already in use here. This needs the composite index plus an Alembic revision, since `migrations/versions/` still ends at `0004`.

One thing to settle together with the ordering question above: the `or_(...)` disjunction at line 187, and particularly the `last_message_at IS NULL` arm, will keep the planner off a clean range scan even once the composite exists. Keying on `COALESCE(last_message_at, created_at)` collapses this to a plain two-term keyset and makes the index usable. If you go that way the index has to be an expression index on the coalesced value, not on the raw column, and the cursor has to encode the coalesced value too, or the predicate and the sort disagree and pages skip rows.

src/agent_manager/domain/models.py
```diff
@@ -56,6 +56,12 @@ class ConversationSession:
     expires_at: datetime | None = None
 
 
+@dataclass(frozen=True)
+class PaginatedSessions:
```
AmitAvital1: This is locked to one entity, so the next table that needs paging gets a copied `PaginatedMessages` with a field named `messages`. A generic costs the same lines and is reusable as it stands:

```python
@dataclass(frozen=True)
class Page[T]:
    items: Sequence[T]
    next_cursor: str | None = None
```

`items` then matches the field name on the wire, which removes the rewrap in `routes/conversations.py:57-65` and lets the API schema be a single `PaginatedResponse[ConversationSummary]` rather than one class per endpoint.

Two smaller points. This belongs in its own `domain/pagination.py` rather than among `ConversationSession` and `Message`; a page is a query-result envelope, not a conversation concept, and it only sits in `domain/` because the port returns it. And `frozen=True` with a mutable `list` field doesn't buy much — `Sequence` or `tuple` would.

src/agent_manager/domain/repository.py
```diff
@@ -66,8 +67,10 @@ async def create_session(
     async def get_session(self, session_id: str) -> ConversationSession | None: ...
 
     @abstractmethod
-    async def list_sessions(self, user_id: str, *, limit: int = 50) -> list[ConversationSession]:
-        """A user's sessions, most-recently-active first."""
+    async def list_sessions(
+        self, user_id: str, *, limit: int = 50, cursor: str | None = None
```
AmitAvital1: `limit` and `cursor` are one concept travelling as two loose parameters through four layers, and each layer restates its own default: `50` here, `50` in `conversation_service.py:140`, `20` in the route, `20` in the widget client, and `20` again as a literal at `AgentChatApp.tsx:221` and `:231`. The service default is unreachable through the API and silently disagrees with it.

```python
@dataclass(frozen=True)
class PageRequest:
    limit: int = DEFAULT_PAGE_LIMIT
    cursor: str | None = None

    def __post_init__(self) -> None:
        if not 1 <= self.limit <= MAX_PAGE_LIMIT:
            raise ValueError(f"limit must be 1..{MAX_PAGE_LIMIT}")
```

The port becomes `list_sessions(user_id, page: PageRequest)`, the service forwards one argument, and the route builds it with `Depends(PageRequest)` so FastAPI still exposes `limit` and `cursor` as query parameters. Adding a `direction` later is one field rather than a signature change in four files.

This also fixes a real gap: the `le=100` cap exists only in the HTTP adapter, so nothing below it bounds `limit` and a direct call goes straight to the database.

```python
await service.list_conversations(caller, limit=10_000_000)
```

If that bound is an invariant of the listing it belongs on the port, not on one of its callers.

tests/agent_manager/test_repository_contract.py
```diff
@@ -56,7 +56,7 @@ async def test_create_session_never_reassigns_an_existing_owner(repo: Repository
     stored = await repo.get_session("shared-id")
     assert stored is not None
     assert stored.user_id == "alice"
-    assert await repo.list_sessions("bob") == []
+    assert (await repo.list_sessions("bob")).sessions == []
```
AmitAvital1: This file is parametrized over both adapters and exists to prove they're interchangeable, but it only received mechanical `.sessions` fixups. The pagination test went into `test_pagination.py`, where `test_sql_repository_pagination_and_ordering` covers `SqlRepository` only, so the memory backend's keyset path has no ordering coverage at all despite being the default.

That matters because the predicate is now written twice, in SQLAlchemy expressions and in Python, with nothing asserting they agree. The timezone divergence on the `sql_repository.py:221` thread is exactly that kind of drift.

The cases worth adding here, running against both adapters:

- walking every page returns each session once, in the same order as an unpaginated read
- sessions sharing a `last_message_at`, so the `session_id` tiebreak is exercised
- rows with `last_message_at IS NULL`, including a page boundary landing on one
- a boundary landing exactly on `limit`, where `next_cursor` must be `None` rather than an empty final page
- a malformed cursor raising the same error from both

Two implementations of one contract is acceptable duplication once CI holds them equal. Without that it's a liability.

src/agent_manager/api/static/widget/react/AgentChatApp.tsx
```diff
@@ -114,6 +114,9 @@ export function AgentChatApp({
   const canStop = isExecutionActive;
 
   const [threads, setThreads] = useState<ThreadSummary[]>([]);
+  const [nextCursor, setNextCursor] = useState<string | null>(null);
+  const [loadingMoreThreads, setLoadingMoreThreads] = useState(false);
+  const [hasMoreThreads, setHasMoreThreads] = useState(false);
```
AmitAvital1: `hasMoreThreads` is always exactly `nextCursor !== null` — both assignments derive it from the cursor they just stored. That's two pieces of state for one fact, updated by separate `setState` calls, so they can disagree.

`const hasMoreThreads = nextCursor !== null` removes the risk and drops it from the dependency array at line 240.

Related, the page size `20` is a bare literal at lines 221 and 231 while `AgentChatClient.listConversations` already defaults to `20` and the route declares `Query(default=20)`. A named constant would keep the initial load and the load-more path from drifting apart.

src/agent_manager/api/static/widget/react/useConversation.ts
```diff
@@ -183,7 +184,11 @@ export function useConversation(
     [client],
   );
 
-  const listThreads = useCallback(() => client.listConversations().catch(() => []), [client]);
+  const listThreads = useCallback(
+    (limit?: number, cursor?: string | null) =>
+      client.listConversations(limit, cursor).catch(() => ({ items: [], next_cursor: null })),
```
AmitAvital1: This catch turns every failure into an empty page. On the initial load that renders "No conversations yet" for a request that actually failed; on a subsequent page it sets `next_cursor: null`, which flips `hasMoreThreads` to `false` permanently, so one transient blip silently truncates the list with no error shown and no way to retry. The user is left believing those conversations no longer exist.

The widget already has what's needed: `AgentChatHttpError` carries `status` and `error_type`, and the other flows surface it. Let the error propagate and have `openThreads` and `loadMoreThreads` render it, so "no more results" and "the request failed" stop looking identical.

src/agent_manager/api/static/widget/react/AgentChatApp.tsx
```diff
@@ -162,8 +165,8 @@ export function AgentChatApp({
 
   const refreshUsage = useCallback(
     async (cid: string) => {
-      const next = await conversation.loadUsage(cid);
-      setUsageById((prev) => ({ ...prev, [cid]: next }));
+      const u = await conversation.loadUsage(cid);
```
AmitAvital1: Unrelated rename (`next` to `u`) in a function this PR doesn't otherwise touch, plus a blank line removed at line 250. Both show up in the bundle diff too. Worth reverting to keep the diff to the change itself.

src/agent_manager/infrastructure/persistence/sql_repository.py
```diff
@@ -151,16 +175,58 @@ async def get_session(self, session_id: str) -> ConversationSession | None:
...
-            rows = (await session.exec(stmt)).all()
-        return [_session(row) for row in rows]
+            rows = list((await session.exec(stmt)).all())
+
+        has_more = len(rows) > limit
+        if has_more:
+            rows = rows[:limit]
+
+        next_cursor = (
+            encode_cursor(rows[-1].last_message_at, rows[-1].session_id)
```
AmitAvital1: `encode_cursor` receives the raw ORM row rather than the `_utc()`-normalized value that `_session()` puts on the domain object, and it calls `.isoformat()` without normalizing. The two adapters therefore emit different formats for the same value:

```
SQL cursor decodes to : datetime(2026, 8, 20, 12, 2)                # naive
SQL domain object has : datetime(2026, 8, 20, 12, 2, tzinfo=utc)    # aware
MEM cursor decodes to : datetime(2026, 8, 20, 12, 2, tzinfo=utc)    # aware
```

Two failures follow.

A naive cursor reaching the in-memory backend raises:

```
TypeError: can't compare offset-naive and offset-aware datetimes
  memory_repository.py:143  sessions = [s for s in sessions if _is_after_cursor(...)]
  memory_repository.py:35   if s.last_message_at < cursor_t:
```

An offset-carrying cursor moves the page boundary on SQLite, because SQLAlchemy's SQLite `DATETIME` bind processor has no timezone field in its format string and drops the offset rather than converting it. Same instant, two encodings:

```
cursor as UTC     -> ['s-b', 's-a']
cursor as +05:00  -> ['s-c', 's-b', 's-a']    # boundary row returned twice
```

A row coming back a second time is the shape of a pagination loop: the client keeps receiving a cursor pointing at a row it already holds. This is reachable as soon as a Postgres-issued cursor, which carries `+00:00` natively, meets a SQLite instance.

Normalize in both directions with `astimezone(UTC)`, and encode from `_session(row).last_message_at` so the cursor and the response body agree on the timestamp. A round-trip test with a naive input and a non-UTC offset would pin it.

tests/e2e/widget.spec.ts
```diff
@@ -31,7 +31,7 @@ async function mockConversationApi(
     await route.fulfill({
       status: 200,
       contentType: "application/json",
-      body: JSON.stringify(options.threads ?? []),
+      body: JSON.stringify({ items: options.threads ?? [], next_cursor: null }),
```
AmitAvital1, line 34: All three mocks in the e2e suite hardcode `next_cursor: null` — here, at line 47, and at line 150 — so the widget always sees one complete page and `loadMoreThreads` is never called.

That leaves the entire new UI path uncovered: the scroll trigger, the append, the dedup, the loading indicator, and the transition to exhausted. These edits keep the suite green without testing what was added, and the widget's own harness in `widget.test.mjs` wasn't extended for `listConversations` either.

One spec whose mock returns a real `next_cursor` on the first call and `null` on the second would cover the flow, and would also exercise the races on the `AgentChatApp.tsx:222` thread.

src/agent_manager/api/schemas.py
```diff
@@ -47,6 +47,11 @@ class ConversationSummary(BaseModel):
     last_message_at: datetime | None = None
 
 
+class PaginatedConversationsResponse(BaseModel):
```
AmitAvital1, line 50: This changes `GET /conversations` from a JSON array to an object, which breaks any consumer that isn't the widget. Together with the `Repository.list_sessions` signature change in `domain/repository.py:70`, that is two contract changes in one PR.

The repository requires an ADR for this:

- `docs/README.md:31` — "adr/ — architecture decision records for contract-level changes"
- `.ai/workflows/code-review.md:38` — "Any contract change is backed by an ADR (or flagged as blocking)."
- `.ai/skills/architecture-review.md`, step 5 — "If the YAML schema, plugin contracts, or API shape changes, require an ADR and assess compatibility."

`docs/adr/` still holds only 0001 and 0002. A `0003-cursor-pagination-and-list-response-envelope.md` should record the envelope break, the cursor format, where empty conversations sort, and the compatibility story. The ordering question in particular is a decision worth making explicitly in the ADR rather than letting it fall out of a sort key.

Two related gaps. `GET /conversations` isn't in `docs/api.mdx` at all, so after this PR the only description of the envelope and the `limit`/`cursor` parameters is the source. And commit `984eb09` is a plain `feat:` with no `!` and no `BREAKING CHANGE:` footer, so release-please will ship a response-shape break as a minor bump.

AmitAvital1 (commented):
Additional notes on the test changes and the response envelope.

rishu685:
Thanks for the thorough review @AmitAvital1! All 20 suggestions have been implemented and verified across the backend, database, frontend widget, and test suite.

### Key Changes Implemented:

1. **Standalone Codec & Layer Decoupling (`pagination.py`)**:
   - Extracted `encode_cursor`, `decode_cursor`, and `InvalidCursorError` into `src/agent_manager/infrastructure/persistence/pagination.py`. Memory and SQL repositories no longer import across infrastructure bounds.
   - Standardized timestamps to timezone-aware UTC (`+00:00`).
   - Narrowed cursor decode error handling strictly to expected exception types `(binascii.Error, UnicodeDecodeError, json.JSONDecodeError, ValueError, KeyError, TypeError)` without logging/leaking raw client input.

2. **Coalesced Keyset Ordering (`COALESCE(last_message_at, created_at)`)**:
   - Updated both `SqlRepository` and `MemoryRepository` to sort by `COALESCE(last_message_at, created_at)`. Newly created empty sessions now appear at the top of the history list instead of sinking to the bottom.

3. **Domain Abstraction & Envelope**:
   - Introduced generic `Page[T]` (`items: list[T]`, `next_cursor: str | None`) with a backwards-compatible `.sessions` property.
   - Created `PageRequest` dataclass containing `DEFAULT_PAGE_LIMIT = 20` and `MAX_PAGE_LIMIT = 100`.

4. **API Error Mapping**:
   - Wrapped `GET /conversations` with `as_http_error()` to map `InvalidCursorError` to HTTP 400 Bad Request `{ "error_type": "invalid_cursor", "message": "invalid pagination cursor" }`.

5. **Database Index & Migration**:
   - Added composite index `idx_conversation_sessions_user_last_message` on `(user_id, last_message_at, session_id)` in `tables.py` and generated Alembic migration `0005_add_session_pagination_index.py`.

6. **React Widget Infinite Scroll Fixes**:
   - Extracted module constants `THREADS_PAGE_SIZE = 20` and `SCROLL_THRESHOLD_PX = 40`.
   - Derived `hasMoreThreads = nextCursor !== null`.
   - Added `isLoadingMoreRef` (`useRef`) synchronous guard to eliminate fast-flick scroll race conditions and duplicate page fetches.
   - Removed `.catch()` swallow in `useConversation.ts` so API errors propagate cleanly.

7. **Documentation & Verification**:
   - Updated `docs/api.mdx` with `GET /conversations` query parameters, envelope structure, and opacity guidelines.
   - Added unit/contract tests for repository contracts, API cursor error responses, and Playwright E2E infinite scroll test coverage.

**Verification Results:**
- `pytest`: 873 passed (0 failures)
- `playwright`: 38 passed (0 failures)

src/agent_manager/api/static/widget/react/AgentChatApp.tsx
```diff
@@ -213,10 +216,29 @@ export function AgentChatApp({
   }, [inline]);
 
   const openThreads = useCallback(async () => {
-    setThreads(await conversation.listThreads());
     setThreadsOpen(true);
+    setLoadingMoreThreads(true);
+    const res = await conversation.listThreads(20, null);
+    setThreads(res.items);
```
AmitAvital1: There's no generation counter or `AbortController` here, and `onClose` doesn't reset anything, so a stale in-flight request can corrupt a freshly reopened drawer.

Scroll deep enough that page 4 is in flight on a slow connection, pick a thread so the drawer closes, then reopen it. `openThreads` resets to page 1, then the page-4 response lands, appends its items after page 1's, and sets `nextCursor` to cursor 4. Pages 2 and 3 are now missing from the list and unreachable by scrolling further. The `existingIds` set at line 233 guards against duplicates; nothing guards against skips.

There's a smaller version of the same problem in `handleScroll` at line 975: both it and the guard at line 229 read state captured in a closure, so two scroll events dispatched before React flushes will both see `false` and both fire. That one is bounded — same cursor, dedup handles it — so it costs a round-trip rather than data.

A request-generation counter checked before each `setState` covers both.

src/agent_manager/api/static/widget/react/AgentChatApp.tsx
```diff
@@ -952,7 +989,7 @@ function ThreadDrawer({
         <SquarePenIcon aria-hidden />
         New chat
       </button>
-      <div className="thread-list">
+      <div className="thread-list" onScroll={handleScroll}>
```
AmitAvital1, line 1034: Loading more is reachable only through `onScroll`. If the first page of 20 doesn't overflow `.thread-list`, no scroll event ever fires and the remaining conversations are unreachable even though `next_cursor` says they exist. That's a tall viewport, a short drawer, or any future page size below what fits on screen.

An `IntersectionObserver` sentinel handles both that case and the normal one. An explicit "Load more" button would also work, and would give the error path somewhere to surface a retry.

src/agent_manager/api/static/widget/api/AgentChatClient.ts
```diff
@@ -67,16 +68,22 @@ export class AgentChatClient {
...
+  async listConversations(limit = 20, cursor?: string | null): Promise<PaginatedThreads> {
+    const params = new URLSearchParams({ limit: String(limit) });
+    if (cursor) params.set("cursor", cursor);
+    const response = await this.request(`/conversations?${params.toString()}`);
 
     const data = await response.json();
-    if (!Array.isArray(data)) return [];
-    return data.map((thread) => ({
+    const rawItems = Array.isArray(data.items) ? data.items : [];
+    const items: ThreadSummary[] = rawItems.map((thread: any) => ({
```
AmitAvital1: `(thread: any)` is the only explicit `any` in the widget source, in a project with `"strict": true` (`tsconfig.widget.json:7`). It compiles, but it severs the compile-time link to `ThreadSummary`, so a server-side rename of `conversation_id` fails silently at runtime instead of failing the build.

Two related lines in the same block. Line 77's `Array.isArray(data.items) ? data.items : []` means an older server still returning a bare array renders as an empty list rather than any signal. And line 85's `data.next_cursor ? String(...) : null` treats `""` as exhausted; `?? null` says what's meant.

AmitAvital1 (commented):
Further notes after a closer pass over the cursor handling, ordering, and the widget scroll path.

AmitAvital1:
@rishu685 Thanks for the write-up, but none of it is on the PR yet, so I want to flag that before you spend more time.

The head is still `d25be318`, the commit I reviewed: 5 commits, 19 files, +434/-72. Your fork's branch is on the same SHA, and both new files 404 at that ref:

```
src/agent_manager/infrastructure/persistence/pagination.py                 -> 404
src/agent_manager/infrastructure/persistence/migrations/versions/0005_*.py -> 404
```

Could you push?

One thing worth ruling out on the test run. Under `uv run pytest` in this repo, `agent_manager` can resolve to an editable install pointing at a different checkout rather than the working tree, in which case the suite passes against pre-PR code. It caught me out during this review. Quick check:

```python
import agent_manager.domain as d
print(d.__file__, hasattr(d, "PaginatedSessions"))
```

If that prints a path outside your tree, force it with `PYTHONPATH=$PWD/src`. For reference, the current PR head is 870 passed, so your 873 is consistent with the tests you describe having been added.

Two things from your description that I'd fix before pushing:

The index won't serve the sort. Sorting on `COALESCE(last_message_at, created_at)` needs an expression index on the coalesced value; a plain index on `(user_id, last_message_at, session_id)` won't be used, so the migration would add an index nothing touches.

The cursor has to encode the coalesced value too. If it still encodes `last_message_at` while the ORDER BY uses `COALESCE`, the predicate and the sort disagree and pagination skips rows. A page walk over data mixing NULL and non-NULL `last_message_at` is the test that catches it.

Also, on `Page[T]` with a backwards-compatible `.sessions` property: this is an internal port with one caller, so there's nothing to stay compatible with. That's the shim worth removing rather than adding.

I've consolidated the inline comments so each finding sits in one place. Push whenever you're ready and I'll review the real diff.

AmitAvital1 (changes requested):
Requesting changes at `d25be318`.

The approach is right and the core keyset logic holds up. Paging both backends across ties, NULLs, and every page size from 1 to 4 produced identical, complete, duplicate-free sequences, `ConversationService` stays a thin pass-through, and the suite is green at 870 passed with lint and mypy clean. Nothing below requires rethinking the design.

Six things to resolve before merge:

1. A malformed cursor returns HTTP 500. This is the only route in `conversations.py` without an error-mapping wrapper, and an empty `?cursor=` reaches the decoder too.
2. Cursors aren't timezone-normalized. A naive cursor raises `TypeError` in the memory backend, and an offset-carrying cursor silently re-emits the boundary row on SQLite.
3. The new query has no supporting index and no migration. `EXPLAIN QUERY PLAN` shows a temp B-tree sort on every page, so the change doesn't yet deliver the performance it exists for.
4. Empty conversations now come back in arbitrary order, since dropping the `created_at` fallback leaves `uuid4().hex` as the effective sort key. The `set()`-based assertion in `test_pagination.py` hides it.
5. `memory_repository.py` imports the cursor codec from `sql_repository.py`. Sibling adapters behind one port shouldn't depend on each other, and it's why the codec has no owner and the two ended up disagreeing on its format.
6. No ADR, for two simultaneous contract changes, against three separate written policies in this repo.

Beyond the fixes, the two design points worth taking are a generic `Page[T]` and a `PageRequest` that owns the page-size default and cap. That's what makes this pagination the repo has rather than pagination this endpoint has, which is the goal in #116.

Most of the test gaps close by moving the pagination cases into `test_repository_contract.py`, where they run against both adapters and would have caught items 2 and 4.

src/agent_manager/domain/__init__.py
```diff
@@ -2,27 +2,39 @@
...
+    Page,
+    PageRequest,
+    PaginatedSessions,
     Role,
     TokenBudgetUsage,
     User,
     thread_title,
 )
 from agent_manager.domain.repository import Repository
+from agent_manager.infrastructure.persistence.pagination import InvalidCursorError
```
AmitAvital1: The sibling-adapter import is gone, but this replaced it with a dependency pointing the wrong way. The line directly above this file's own docstring reads "Domain layer: value objects and ports. Pure Python, no frameworks."

Before this PR neither `domain` nor `application` imported `infrastructure` anywhere. There are now two:

```
domain/__init__.py:22       from agent_manager.infrastructure.persistence.pagination import InvalidCursorError
application/errors.py:3     from agent_manager.infrastructure.persistence.pagination import InvalidCursorError
```

`domain` is the innermost layer and defines the `Repository` port; having it reach outward into an adapter package inverts the direction the port exists to enforce.

Nothing breaks at runtime today — `pagination.py` is stdlib-only, and I confirmed `import agent_manager.domain` doesn't pull SQLAlchemy — so this is a design problem rather than a live one.

An invalid cursor is a domain concept, not a storage detail. Move `InvalidCursorError` into `domain/models.py` beside `PageRequest` and `Page`, and have `infrastructure/persistence/pagination.py` import it from there. Both lines above then disappear.

docs/api.mdx
```diff
@@ -166,6 +166,36 @@ Response:
 
 Errors: `409` the supplied `session_id` belongs to another user.
 
+### `GET /conversations`
```
AmitAvital1, line 169: Good to have this documented, and the opaque-cursor note is the right thing to say.

Two things still outstanding on the contract side.

There's no ADR. `docs/adr/` still holds only 0001 and 0002, and this PR changes two contracts at once: the `Repository` port signature and the response shape of this endpoint. The requirement is written in three places (`docs/README.md:31`, `.ai/workflows/code-review.md:38`, `.ai/skills/architecture-review.md` step 5). A `0003-cursor-pagination-and-list-response-envelope.md` recording the envelope change, the cursor format, and the decision to order on `COALESCE(last_message_at, created_at)` would close it. That last one is worth writing down — it is a deliberate choice now, and the next person changing the sort key should find the reasoning rather than infer it.

And the array-to-object change still ships as a minor release. `984eb09` is a plain `feat:` and `3b2c7b0` is a `fix:`, neither carrying a `!` or a `BREAKING CHANGE:` footer.

src/agent_manager/api/static/widget/react/AgentChatApp.tsx
```diff
@@ -213,10 +221,49 @@ export function AgentChatApp({
...
+    const currentGen = threadsGenerationRef.current;
     setThreadsOpen(true);
+    setLoadingMoreThreads(true);
+    isLoadingMoreRef.current = true;
+    try {
+      const res = await conversation.listThreads(THREADS_PAGE_SIZE, null);
+      if (threadsGenerationRef.current !== currentGen) return;
+      setThreads(res.items);
+      setNextCursor(res.next_cursor);
+    } finally {
```
AmitAvital1, line 240: The swallow in `useConversation` is gone, which is the right first half. But `openThreads` and `loadMoreThreads` are `try`/`finally` with no `catch`, and both are invoked as `void openThreads()` (line 618) and `void loadMoreThreads()` (line 649), so a rejected request becomes an unhandled promise rejection and the user still sees nothing.

It is better than before — `setThreads` never runs, so the list keeps its contents, and `nextCursor` is untouched, so scrolling again retries. But "the request failed" and "there is nothing here" still look identical.

`AgentChatHttpError` already carries `status` and `error_type`. A `catch` that stores the message and a line in the drawer to render it would finish this, and would give the case below somewhere to offer a retry.

src/agent_manager/api/static/widget/react/AgentChatApp.tsx
```diff
@@ -929,17 +978,30 @@ function ThreadDrawer({
...
+  loadingMore: boolean;
+  hasMore: boolean;
+  onLoadMore: () => void;
   onSelect: (conversationId: string) => void;
   onNew: () => void;
   onClose: () => void;
 }) {
+  const handleScroll = (e: React.UIEvent<HTMLDivElement>) => {
+    const { scrollTop, clientHeight, scrollHeight } = e.currentTarget;
+    if (scrollHeight - scrollTop - clientHeight < SCROLL_THRESHOLD_PX && hasMore && !loadingMore) {
```
AmitAvital1: Still the only path to loading more. If the first page of 20 does not overflow `.thread-list`, no scroll event ever fires and the remaining conversations stay unreachable even though `next_cursor` says they exist. A tall viewport is enough.

An `IntersectionObserver` on a sentinel element covers this and the normal case together. A "Load more" button would too, and would pair with the error surface above.

tests/agent_manager/test_repository_contract.py
```diff
@@ -468,18 +474,49 @@ async def test_linking_a_visitor_moves_their_sessions_once(repo: Repository) ->
...
-    assert await repo.list_sessions("ext:bob") == []
+    assert (await repo.list_sessions("ext:bob")).items == []
 
 
 async def test_linking_an_unknown_visitor_is_a_no_op(repo: Repository) -> None:
     await repo.upsert_user("ext:alice")
     assert await repo.link_anonymous_user("anon:never-seen", "ext:alice") == 0
+
+
+async def test_pagination_contract(repo: Repository) -> None:
```
AmitAvital1, line 494: This is the right home for these, and running the walk, the exact-limit boundary and the malformed cursor across both adapters covers most of what matters.

One case is missing here, and it happens to be the one this round was about. All three sessions are created without messages, so every row falls back to `created_at` and the dated/undated boundary is never crossed. The mixed case is exercised well in `test_pagination.py`, but only against `SqlRepository` — which leaves the `COALESCE` behaviour of the default backend unpinned.

Extending this to five sessions, three with a `last_message_at` (two of them sharing a value, so the `session_id` tiebreak is exercised) and two without, then asserting the full sequence across pages, would hold both implementations to the same ordering.

src/agent_manager/domain/models.py
```diff
@@ -56,6 +58,31 @@ class ConversationSession:
...
+    def __post_init__(self) -> None:
+        bounded_limit = max(1, min(self.limit, MAX_PAGE_LIMIT))
+        object.__setattr__(self, "limit", bounded_limit)
+        if self.cursor is not None and not self.cursor.strip():
+            object.__setattr__(self, "cursor", None)
+
+
+@dataclass(frozen=True)
+class Page(Generic[T]):
+    items: list[T]
```
AmitAvital1: Minor: `frozen=True` with a mutable `list` field doesn't give you much — the reference is frozen, the contents aren't. `Sequence[T]` in the annotation, or `tuple[T, ...]` if you want it enforced, would match the intent.

AmitAvital1 (changes requested):
This is a big step and most of it landed properly. I re-ran everything rather than going by the description:

```
malformed cursor   500 plain text        ->  400 {"error_type": "invalid_cursor", ...}
cursor payload     '2026-08-20T14:00:00' ->  '2025-12-22T00:00:00+00:00'
ordering           [s0, s1, s2, new]     ->  [new, s0, s1, s2]
query plan         USE TEMP B-TREE       ->  USING INDEX ... (user_id=? AND <expr><?)
```

The index is an expression index matching the `COALESCE` sort, with migration `0005`, and page N is a real range seek rather than a sort. Both backends now derive ordering and the keyset predicate from one definition. 875 passed, ruff and mypy clean.

Also resolved: `Page[T]` and `PageRequest` with single constants, a contract test over both adapters, an e2e spec that actually walks a second page, the API docs, and the widget work — generation counter plus a synchronous ref for the races, `any` removed, `hasMoreThreads` derived, and the unrelated churn reverted. Replacing the `.sessions` shim with a plain type alias was the better call.

Two things left before I can approve:

1. `domain/__init__.py` and `application/errors.py` now import from `infrastructure`. The sibling-adapter problem is fixed, but the replacement points the dependency outward from the innermost layer. One file move.
2. Still no ADR, and the response-shape break still has no `BREAKING CHANGE:` footer.

The rest are small and inline.

src/agent_manager/infrastructure/persistence/pagination.py
```diff
@@ -0,0 +1,18 @@
...
+
+from agent_manager.domain.pagination import (
+    InvalidCursorError,
+    decode_cursor,
+    encode_cursor,
+    ensure_utc,
+)
+
+# Backward-compatibility private alias
+_utc = ensure_utc
```
AmitAvital1, line 11: This alias has no callers — everything uses `ensure_utc` directly.

More broadly, this whole module is now a re-export of `domain.pagination`, and the only thing importing it is `tests/agent_manager/test_pagination.py:16`. Both adapters already import from `domain.pagination`.

Pointing that test at `agent_manager.domain.pagination` lets the file be deleted. Same reasoning as dropping the `.sessions` shim earlier: nothing outside the repo depends on this path, so there's no compatibility to preserve.

docs/adr/0003-chat-history-pagination-and-keyset-cursor.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Contract changes
+
+- **BREAKING CHANGE**: `GET /conversations` response shape changed from a flat list `[ConversationSummary, ...]` to a paginated envelope object `{ "items": [ConversationSummary, ...], "next_cursor": "..." | null }`.
+- `GET /conversations` accepts optional query parameters `limit` (integer, 1..100) and `cursor` (opaque string).
+
+## Consequences
+
+- Clients fetch subsequent pages using `next_cursor` until `next_cursor` is `null`.
+- Keyset range seeks eliminate `OFFSET` database performance degradation and prevent skipped/duplicated sessions when thread activity changes mid-page.
```
AmitAvital1, line 38: This claims more than keyset can deliver. Because the sort key is mutable, a thread that receives a message while someone is paging jumps ahead of their cursor and is skipped on the page they were about to load. That's inherent to keyset over a mutable column, and it's the reason the next line says frontends deduplicate — if skips and duplicates were prevented, no dedup would be needed.

Worth stating accurately, since the ADR is what the next reader will trust:

> Keyset range seeks eliminate `OFFSET` performance degradation and keep page boundaries stable against inserts. Because `last_message_at` is mutable, a thread that becomes active mid-scroll can move across page boundaries; the listing is a snapshot from when paging began, and clients deduplicate by `conversation_id`.

Matches the comment already in `loadMoreThreads`.

AmitAvital1 (approved):
Approving. Everything I raised is addressed, and I re-verified the behaviour rather than reading the diff alone:

```
malformed cursor   ->  400 {"error_type": "invalid_cursor", "message": "invalid pagination cursor"}
cursor payload     ->  {"t": "2025-12-22T00:00:00+00:00", "id": "s0"}
ordering           ->  ['new', 's0', 's1', 's2']   full walk, no gaps, no duplicates
query plan         ->  SEARCH ... USING INDEX idx_conversation_sessions_user_active_session
                       (user_id=? AND <expr><?)
```

875 passed, ruff, mypy and `generate-check` all clean.

The layering fix is the right one: moving the whole codec into `domain/pagination.py` puts the cursor contract where both adapters can depend on it inward, rather than shuffling the violation somewhere else. ADR 0003 is more thorough than I asked for — recording why offset was rejected and why the sort key is `COALESCE(last_message_at, created_at)` is exactly what a future reader will need, and the `BREAKING CHANGE:` footer means the release will say so too. The contract test now runs five sessions with mixed dated and never-messaged rows and a shared timestamp for the tiebreak, across both adapters, which is what makes the two keyset implementations trustworthy rather than coincidentally equal.

The drawer error notice with a Retry button is a better answer than what I suggested, and the "Load more" button resolves the no-overflow case at the same time.

Two small things inline, neither blocking. Happy for this to go in with or without them.

Nice work — this moved a long way in a short time.

AmitAvital1 (approved):
Re-approving against `df28800`, since my previous approval landed on `b5e75783` a moment after you'd already pushed past it.

Worth recording why that matters: `b5e75783` did not pass this project's typecheck gate. I ran `mypy src` when the Makefile runs `mypy $(SRC) $(TESTS)`, so I missed three errors in `test_repository_contract.py` — the `Sequence` concatenation and the `_sessions` access on the base `Repository` type. You'd caught and fixed them in `a8f8352` before I even submitted. My verification was narrower than the gate; that's on me, and the approval you had was on a commit CI would have rejected.

Re-ran the full gate on `df28800`:

```
ruff format --check   322 files already formatted
ruff check            All checks passed!
mypy src tests        Success: no issues found in 275 source files
pytest                875 passed
generate-check        clean
```

And re-confirmed the behaviour: malformed cursor returns `400 {"error_type": "invalid_cursor"}`, the cursor payload carries `+00:00`, and a full page walk returns `['new', 's0', 's1', 's2']` with no gaps or duplicates.

The two notes from my last review are still open and still non-blocking: the `infrastructure/persistence/pagination.py` re-export shim, and the "prevent skipped/duplicated sessions" line in the ADR's consequences. Fine either way.

PR #135, opened 2026-09-01
Feat/mcp structured tool results

Asaf-prog:
## Summary

Preserve structured MCP tool results throughout Extra's execution pipeline instead of collapsing every successful tool result into plain text.

This builds on #133, which fixed MCP text extraction, and adds a provider-agnostic normalized tool-result model that keeps:

- model-facing text;
- machine-readable structured data;
- bounded artifact metadata.

The structured result now survives hooks, persistence, idempotent replay, and HITL resume without leaking provider-specific LangChain objects into the core runtime.

## What changed

- Added an immutable normalized tool-result abstraction owned by Extra.
- Preserve MCP `structuredContent` instead of dropping it.
- Preserve local structured `dict` / `list` tool results where applicable.
- Use deterministic canonical JSON for structured-only model-facing fallback.
- Keep only text in the model conversation; structured/artifact data is not automatically exposed through `ToolMessage`.
- Added versioned serialization with backwards compatibility for legacy text-only execution records.
- Preserve structured results through idempotent replay and HITL resume.
- Added atomic claim/wait semantics so concurrent duplicate calls do not execute the same tool side effect more than once.
- Keep existing text-based `transform_tool_result` hooks compatible while preserving structured data.
- Added bounded artifact normalization and safer handling of malformed/non-serializable provider values.
- Avoid logging raw structured tool results or provider-supplied values.

## Runtime flow

Provider / LangChain result
        ↓
Tool-result normalization
        ↓
NormalizedToolResult
├── text
├── structured
└── bounded artifact metadata
        ↓
Hooks
        ↓
Execution ledger
        ↓
Replay / HITL resume

text only
        ↓
Model conversation

## Backwards compatibility

- Existing local string tools behave as before.
- MCP text-only results keep the behavior introduced by #133.
- Legacy text-only execution records can still be restored.
- Existing text transformation hooks remain supported.

## Tests

Added/updated coverage for:

- MCP text-only and multi-block results;
- MCP `structuredContent`;
- structured-only results;
- deterministic JSON fallback;
- local structured and string tool results;
- malformed/non-serializable provider values;
- artifact size and nesting limits;
- sequential and concurrent idempotent replay;
- terminal execution-state protection;
- legacy ledger compatibility;
- structured result preservation through hooks;
- HITL suspend/resume;
- prevention of structured data leaking into logs/callback metadata.

## Validation

- Ruff format/lint passed.
- Mypy passed.
- Pytest: 965 passed.
- `git diff --check` passed.

## Out of scope

- Full multimodal MCP support.
- Binary artifact persistence.
- Widget rendering for structured/artifact results.
- Universal output-schema validation.
- Provider-specific model serialization strategies.


src/agent_engine/engine/langgraph/tools/tool_invoker.py
```diff
@@ -302,16 +312,24 @@ async def _record_error(self, call: _ToolCall, exc: Exception, latency_ms: int)
...
+        try:
+            await self._hook_manager.run_on_tool_error(
+                current_run_context.get(),
+                self._call_context(call, "failed", latency_ms, error=error),
+            )
+        except BaseException:
+            await self._publish_processing_failure(call, "Tool error processing failed")
+            raise
+        result = NormalizedToolResult.text_only(model_text or f"Tool error: {exc}")
+        await self._execution_manager.finish_execution(call.exec_id, status="failed", result=result)
```
AmitAvital1: Cancellation while awaiting this write leaves the claim at `started`: `_record_error()` is outside the recovery wrapper used by `_record_success()`.
Please shield/finalize it and add a cancellation test; otherwise duplicate callers can wait forever.

src/agent_engine/engine/langgraph/tools/tool_result_normalizer.py
```diff
@@ -0,0 +1,225 @@
...
+        self.string_chars_left -= size
+        return True
+
+
+def normalize_tool_result(result: object) -> NormalizedToolResult:
+    """Normalize one LangChain or direct local-tool result exactly once."""
+    try:
+        if isinstance(result, ToolMessage):
+            model_text, content_structured = _tool_message_content(result.content)
+            artifact_structured, artifact = _split_artifact(result.artifact)
```
AmitAvital1, line 63: This branch relies on the artifact contract introduced by `langchain-mcp-adapters` 0.2.0, but `pyproject.toml` still permits 0.1.x.
Could we bump the minimum version? Otherwise valid installations silently lose `structuredContent`, and older artifact shapes can fail normalization.

src/agent_engine/runtime/tool_results.py
```diff
@@ -0,0 +1,167 @@
...
+        structured: object | None = None,
+        artifact: object | None = None,
+    ) -> None:
+        if not isinstance(text, str):
+            raise ToolResultValidationError("tool result text must be a string")
+        object.__setattr__(self, "text", text)
+        object.__setattr__(
+            self,
+            "_structured_json",
+            _canonical_json(structured, field_name="structured result")
```
AmitAvital1, line 81: `structured` is canonicalized and later decoded, copied, and persisted with only a depth check; unlike `artifact`, there is no item or byte budget.
A large MCP response can multiply memory use across normalization, hooks, and replay—please reuse a bounded-JSON policy here.

src/agent_engine/approvals/models.py
```diff
@@ -17,10 +17,13 @@
...
 from dataclasses import dataclass, field
 from enum import StrEnum
-from typing import Any
+from typing import Any, Literal
 
 from agent_engine.approvals.errors import InvalidStateTransition
 from agent_engine.runtime.tool_models import ToolProviderName
+from agent_engine.runtime.tool_results import PersistedToolResult
+
+ToolExecutionStatus = Literal["started", "succeeded", "failed"]
```
AmitAvital1: Could this be a `StrEnum`, consistent with `RunStatus` and `ApprovalStatus` below?
The raw values are repeated in the repository, manager, and invoker; enum members would remove these magic state strings.

AmitAvital1 (changes requested):
Requesting changes for correctness and contract issues found in the structured-result execution path; details are inline.

Asaf-prog:
Thanks for the detailed review — I addressed all four findings:
- Made terminal execution writes cancellation-safe. Failed results are persisted before secondary usage/hook processing, and cancellation waits for the shielded ledger write to finish. Added a regression proving concurrent duplicate callers do not hang or re-execute the provider.
- Raised the langchain-mcp-adapters minimum version to >=0.2, matching the structured-content artifact contract used by the normalizer.
- Added bounded JSON validation for structured results: depth, total values, string/key bytes, integer size, and final encoded size are limited. Oversized MCP structured results now produce controlled tool errors.
- Replaced repeated execution-state strings with the wire-compatible ToolExecutionStatus StrEnum.
Documentation and ADR 0004 were updated accordingly. The full repository gate passes: 969 tests, Ruff, formatting, and mypy

src/agent_engine/engine/langgraph/tools/tool_invoker.py
```diff
@@ -255,63 +223,114 @@ async def _cached_result(self, call: _ToolCall) -> str | None:
...
+                exc,
+                _elapsed_ms(start),
+                model_text="Tool error: invalid tool result",
+            )
 
-    async def _record_error(self, call: _ToolCall, exc: Exception, latency_ms: int) -> str:
+        try:
+            return await self._record_success(call, normalized, _elapsed_ms(start))
+        except BaseException:
+            await self._publish_processing_failure(call, "Tool result processing failed")
```
AmitAvital1: Cancellation during this successful terminal write is re-raised after the ledger is already persisted as `SUCCEEDED`; this handler then attempts a second terminal write as `FAILED`.
I reproduced `ValueError: ... already terminal` instead of `CancelledError`—please avoid republishing here and add the symmetric success-path cancellation test.

AmitAvital1 (changes requested):
Re-reviewed `91f90c3`: the dependency floor, bounded JSON policy, and execution-status enum are fixed and those threads are resolved.
Cancellation safety still double-finalizes a successful execution under cancellation; see the inline reproduction, so I cannot approve this head yet.

Asaf-prog:
Thanks for the re-review and clear reproduction. I’ve fixed the remaining cancellation race.

The success path now distinguishes between:

- failures during pre-publication result processing, which finalize the execution as `FAILED`; and
- cancellation during the terminal success write, which preserves the committed `SUCCEEDED` result and propagates `CancelledError` without attempting a second write.

I also added the symmetric regression test. It verifies that cancellation during a blocked successful write:

- does not raise the terminal-overwrite `ValueError`;
- preserves the successful result;
- unblocks the duplicate caller;
- returns the same result to that caller; and
- executes the provider exactly once.

Validation is clean: Ruff, mypy, and the full suite pass with **970 tests passed**.

AmitAvital1 (approved):
Re-reviewed `828b6de`: the cancellation race now preserves the terminal success result, propagates `CancelledError`, and keeps provider execution exactly once.
The regression test and independent reproduction pass; all previous findings are resolved and CI is green.

